PR #4, opened 2015-02-12
distinguish msg from docker logs

jhadvig:
Just a small change to distinguish building/testing message from docker logs, which looks like:
```
hack/test.sh rhel7 
---> Building nodejs-0.10-rhel7
Sending build context to Docker daemon 28.67 kB
Sending build context to Docker daemon 
Step 0 : FROM rhel7
 ---> 8a06c82684c8
```


jhadvig:
@soltysh thoughts ? :)

soltysh:
Yep, that makes sense. LGTM.

PR #10, opened 2015-02-17
Baking STI scripts into the image

jhadvig:


jhadvig:
[test]

0.10/Dockerfile
```diff
@@ -41,4 +38,4 @@ USER nodejs
 
 EXPOSE 3000
 
-CMD ["/opt/nodejs/bin/usage"]
\ No newline at end of file
+CMD ["/usr/local/bin/usage"]
```
mfojtik: This should be /usr/local/sti/ ?

Also if you add this into PATH, you don't have to write absolute path.

bparees:
lgtm

bparees:
appears our test bot is letting us down again.

jhadvig:
@bparees @mfojtik I've updated the PR with commit that adds an extra test case, which tests if usage script path is set correctly in the Dockerfile.
Thoughts ?

mfojtik:
@jhadvig can you get rid of ENV APP_ROOT . altogether?

mfojtik:
@jhadvig also if the WORKDIR is set to application runtime directory, make sure you nuke all absolute paths from all STI scripts.

bparees:
@jhadvig new test case looks good

jhadvig:
@bparees @mfojtik I've updated the commits and also add one which replaces `centos7` image with `openshift/base-centos7`.
Can you please take a look ?

bparees:
lgtm, but reminds me that sooner or later we need a rhel7 base.

mfojtik:
LGTM

PR #18, opened 2015-02-20
Remove STI_ environment variables from README

mfojtik:
This ENV vars are coming from the base image and they should be explained there. The STI_LOCATION is not needed anymore. The README should have only informations useful for the consumers of the language. It should mention language specific ENV vars that users can use to tweak the behavior of the image.


mfojtik:
@soltysh PTAL

soltysh:
LGTM

PR #25, opened 2015-03-24
Updated sample-app link

soltysh:


README.md
```diff
@@ -48,19 +49,19 @@ you can omit this parameter.**
...
 ---------------------
-To build simple [nodejs-echo-app](https://github.com/ryanj/node-echo) application,
+To build simple [nodejs-sample-app](https://github.com/openshift/sti-nodejs/tree/master/0.10/test/test-app) application,
 using standalone [STI](https://github.com/openshift/source-to-image) and then run the
 resulting image with [Docker](http://docker.io) execute:
 
 *  **For RHEL based image**
     ```
-    $ sti build https://github.com/ryanj/node-echo.git openshift/nodejs-010-rhel7 nodejs-echo-app
+    $ sti build https://github.com/openshift/sti-nodejs.git --contextDir=/0.10/test/test-app/ openshift/nodejs-010-rhel7 nodejs-sample-app
```
mfojtik: omg this is ugly :D 

why not just store the STI config? `--use-config` and then all you need is to execute "sti build --use-config" ?

jhadvig:
LGTM

mfojtik:
not-LGTM ;-) I think we should present cleanest examples possible and I'm not considering the ugly long command line as 'clean'.

README.md
soltysh: With config you'll end up telling user to be in that and not in other directory, with this I don't care if he has or not the source code of this repo. It's more universal this way.

mfojtik: why you care? you just want to build the application with this image, you don't care about all the ugly options to the sti command. The example will be then:

```
$ cd repo
$ sti build --use-config
$ docker run
```

jhadvig: Could it contain both option, like I would understand that someone would care about what are they actually running, but if not, they will have the second shortcut

mfojtik: @jhadvig if they care, then they have STI repo they can go read about STI. As end-user, less you need to type, better.

mfojtik: @jhadvig remember, this is **not** STI repo, just image to happen to be compatible with STI build that we want to present in cleaner way possible :-)

soltysh: I'll be changed when https://github.com/openshift/source-to-image/issues/158 will be resolved.

mfojtik:
reverting my not-LGTM to LGTM as we discussed this issue and created issue: https://github.com/openshift/source-to-image/issues/158

bparees:
lgtm but merge build is failing.

soltysh:
Let's re-[test]

jhadvig:
@soltysh I've already sent a PR in `vagrant-openshift` repo ro fix this issue, waiting for merge

jhadvig:
[test]

jhadvig:
LGTM, fixed the Jekins job.
Merging

PR #28, opened 2015-04-03
Enable scl globally

jhadvig:
@soltysh PTAL

### DIFF
--- a/0.10/.sti/bin/assemble
+++ b/0.10/.sti/bin/assemble
@@ -1,11 +1,13 @@
-#!/bin/bash -e
+#!/bin/bash
 
 # Prevent running assemble in builders different than official STI image.
 # The official nodejs:0.10-onbuild already run npm install and use different
 # application folder.
 [ -d "/usr/src/app" ] && exit 0
 
-source "../lib/nodejs_context"
+source scl_source enable nodejs010
+
+set -e
 
 if [ "$1" = "-h" ]; then
   exec /usr/local/sti/usage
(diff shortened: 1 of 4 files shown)


soltysh:
You should also remove `0.10/contrib/lib/nodejs_context` file.

soltysh:
Other than that LGTM.

0.10/.sti/bin/assemble
```diff
@@ -1,11 +1,13 @@
-#!/bin/bash -e
+#!/bin/bash
 
 # Prevent running assemble in builders different than official STI image.
 # The official nodejs:0.10-onbuild already run npm install and use different
 # application folder.
 [ -d "/usr/src/app" ] && exit 0
 
-source "../lib/nodejs_context"
+source scl_source enable nodejs010
```
mfojtik: why you need to do this when you enable it in bashrc?

0.10/.sti/bin/run
```diff
@@ -1,4 +1,8 @@
-#!/bin/bash -e
+#!/bin/bash
+
+source scl_source enable nodejs010
```
mfojtik: same here :)

0.10/.sti/bin/assemble
jhadvig: bashrc is not loaded in the STI scripts, only when enter the container

jhadvig:
@soltysh updated :) thanks for noticing, guess I overlooked it due to the hunger :hamburger:

0.10/.sti/bin/assemble
mfojtik: source it then

soltysh: Not true, I'm not sourcing it anywhere in [sti-python](https://github.com/openshift/sti-python), `.bashrc` is loaded upon all `bash` invocations and that's what STI is doing.

soltysh:
LGTM.

mfojtik:
@soltysh I suggested to source it to avoid having scl commands and logic all over the place.... we should enable it in one place and then just source it (bashrc).

mfojtik:
@soltysh @jhadvig my point is to allow disabling scl easy (if you want to). in that case I have to comment it just in one place and not in all scripts.

soltysh:
I misread that, but yeah that'd be good. You're welcome to create follow up
PRs for all the images :D

PR #29, opened 2015-04-07
Sourcing .bashrc

jhadvig:
Enable the SCL's by sourcing `.bashrc` as suggested by @mfojtik in the last PR.
@mfojtik PTAL


soltysh:
LGTM.

PR #40, opened 2015-05-29
Updated docker-scripts

soltysh:
Waiting for https://github.com/openshift/sti-base/pull/31


bparees:
what in this is waiting for https://github.com/openshift/sti-base/pull/31 ?

soltysh:
I wanted them to be rebuilt at once, already using the new base image with
labels.

bparees:
eh, your call but builds are cheap and your time babysitting PRs is not :)

PR #53, opened 2015-06-30
Restore old labels for RHEL7 backward compatibility

mfojtik:


mfojtik:
@bparees PTAL

bparees:
lgtm other than my same question as the rest...why only rhel.

PR #54, opened 2015-07-08
Remove contrib/src

rhcarvalho:
Now contrib files are under contrib/etc.

Related to https://github.com/openshift/sti-python/issues/18

@bparees @mfojtik PTAL


bparees:
lgtm.

PR #60, opened 2015-08-03
Retag candidate image after the tests pass

jhadvig:
@mfojtik PTAL


hack/build.sh
```diff
@@ -67,6 +69,11 @@ for dir in ${dirs}; do
 
   if [[ -v TEST_MODE ]]; then
     IMAGE_NAME=${IMAGE_NAME} test/run
+
+    if [[ $? -eq 0 ]] && [[ "${TAG_ON_SUCCESS}" == "true" ]]; then
+      echo "-> Re-tagging ${IMAGE_NAME} image to ${IMAGE_NAME%"-candidate"}"
+      docker tag $IMAGE_NAME ${IMAGE_NAME%"-candidate"}
```
bparees: backwards?

jhadvig:
LGTM

hack/build.sh
jhadvig: So the
- `$IMAGE_NAME` - openshift/nodejs-010-centos7-candidate
- `${IMAGE_NAME%"-candidate"}` - openshift/nodejs-010-centos7

bash black magic :sparkles:

bparees: yup, got it. thanks :)

PR #65, opened 2015-09-04
sti -> s2i rebranding

soltysh:
@bparees I hope that reminds you to do the releaese :wink:


mfojtik:
@soltysh @bparees we have to make sure the usage we print for images is doable for the RHEL7 users. Make sure the binary we package for RHEL7 is called `s2i` and `atomic build` or something. 

@hhorak FYI.

soltysh:
If they use almost released (@bparees looking at your again :wink:) binary there won't be a problem 'cause we produce `s2i` and `sti` binaries during that process.

soltysh:
This covers https://github.com/openshift/source-to-image/issues/275. I'm linking other image PRs with similar changes:
https://github.com/openshift/sti-perl/pull/46
https://github.com/openshift/sti-php/pull/55
https://github.com/openshift/sti-python/pull/61
https://github.com/openshift/sti-ruby/pull/69
https://github.com/openshift/sti-wildfly/pull/57

bparees:
[test]

PR #69, opened 2015-09-29
use env variable for script copy location

bparees:


jhadvig:
Should we also update this https://github.com/openshift/sti-nodejs/blob/master/0.10/s2i/bin/assemble#L11 ?

bparees:
yes, thanks.  updated.

mfojtik:
:+1:

jhadvig:
LGTM

PR #72, opened 2015-10-27
Syncing nodejs image build logic with the rest of the images

jhadvig:
@bparees @praiskup PTAL

I've also incorporated the 'using the whole commit ID change' into this PR.


bparees:
lgtm

PR #74, opened 2015-11-25
Preliminary PR for DEVEX-57

Ian-Stewart-Binks:
@gorkem Can you take a look at this PR? This is only for [DEVEX-57](https://issues.jboss.org/browse/DEVEX-57) (DEVEX-43). 

Notes:
* Naming the directory/image 0.20 may be incorrect. It appears that 0.10 may have been named according to the Node.js version that it supports. The [sti-ruby](https://github.com/openshift/sti-ruby) repository uses 2.0 and 2.2, corresponding to ruby versions. It may be best to pull this into 0.10.
* The env variable for development mode has been intentionally chosen to be NODE_ENV. This limits modularity: if we want to do something in development mode and not produce other side effects that NODE_ENV produces, we are unable to. However, the question is whether or not we want to allow the two to diverge. Any side effects that the NODE_ENV produces, or could conceivably produce in the future, are side effects that we want, e.g. not copying devDependencies under production. Note, however, that NODE_ENV must be set to production. If NODE_ENV is also ever deprecated, we look slightly messy if we don't also keep up with the NODE_ENV replacement.
* I'm deciding to not link to a blog in the documentation in the repository. No real reason other than the blog hasn't yet been written.
* The `run` script doesn't read in the default values from the image. Should it? Is it even possible? the run script would then need to know the name of the image. That could be a bit messy.

### DIFF
--- a/0.20/Dockerfile
+++ b/0.20/Dockerfile
@@ -1,0 +1,41 @@
+FROM openshift/base-centos7
+
+# This image provides a Node.JS environment you can use to run your Node.JS
+# applications.
+
+MAINTAINER SoftwareCollections.org <sclorg@redhat.com>
+
+EXPOSE 8080
+
+ENV NODEJS_VERSION 0.10
+
+LABEL io.k8s.description="Platform for building and running Node.js 0.10 applications" \
+      io.k8s.display-name="Node.js 0.10" \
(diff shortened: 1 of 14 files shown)


Ian-Stewart-Binks:
Whoops. Meant to pull this into my own repository. Closing and redoing.

PR #76, opened 2015-12-07
Change Centos rpms to point to centos hosted collections instead of s…

Ian-Stewart-Binks:
…oftwarecollections.org

From email on scl mailing list:
> Since the website softwarecollections.org has experienced troubles these 
days, I believe we can officially suggest using builds from CentOS Build 
System as the solution -- not only for now, but also for the future, 
because it would happen anyway.

Signed-off-by: Ian Stewart-Binks <istewart@redhat.com>

### DIFF
--- a/0.10/Dockerfile
+++ b/0.10/Dockerfile
@@ -14,9 +14,7 @@ LABEL io.k8s.description="Platform for building and running Node.js 0.10 applica
       io.openshift.expose-services="8080:http" \
       io.openshift.tags="builder,nodejs,nodejs010"
 
-RUN yum install -y \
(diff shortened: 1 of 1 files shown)


bparees:
[test]

bparees:
@Ian-Stewart-Binks any chance you want to make this change in our other image repos also?

bparees:
@Ian-Stewart-Binks nevermind, i'm on it.

Ian-Stewart-Binks:
@bparees sorry about my late reply. I was sick and away from work yesterday. I see you've already switched everything over. Sorry I couldn't be of more help.

PR #78, opened 2015-12-16
Add support for REGISTRY parameter

mfojtik:


bparees:
lgtm.

bparees:
As an aside, this rebuild of nodejs should pick up the fix for the npm lock bug, according to @eliskasl

PR #83, opened 2016-01-25
set npm proxy from http_proxy env

bparees:


bparees:
@mfojtik ptal

jhadvig:
LGTM :+1: 
... just one tiny proposal/nit, wouldn't it be also good to print the set http/https proxy address ?

bparees:
@jhadvig yeah maybe. i did debate that :)

jhadvig:
I see that the outcome of the debate was against the printing :-]

bparees:
@jhadvig added.

jhadvig:
ta ta ta taaa... I'm lovin' it :hamburger:

Issue #87, opened 2016-03-19
add $HOME/node_modules/.bin to the PATH


ryanj:
This should provide a fix for #82

elyscape:
This shouldn't be necessary. npm adds this to the PATH when executing the scripts in package.json.

ryanj:
@elyscape the use case I'm trying to solve is to provide CLI access to scripts - so users won't need to use the "global" installation mode ("npm i -g grunt-cli" or "npm i --global gulp-cli") in order to make grunt, gulp, or bower available to their shell.

elyscape:
Gotcha. That makes sense.

bparees:
fixed by https://github.com/openshift/sti-nodejs/pull/88

PR #88, opened 2016-03-20
Adding ./node_modules/.bin to the $PATH to provide CLI access for npm scripts

ryanj:
For #87 (Needs testing) - I tried to follow the [build instructions](https://github.com/openshift/sti-nodejs#installation), but couldn't get them to work: (`lstat .sti/bin/: no such file or directory; hack/common.mk:21: recipe for target 'build' failed`)

This change *should* give users a way to make scripts "globally available" (CLI accessible)
Without requiring them to use admin-level system functions, like npm's "--global" installation mode.

### DIFF
--- a/0.10/Dockerfile
+++ b/0.10/Dockerfile
@@ -7,7 +7,10 @@ MAINTAINER SoftwareCollections.org <sclorg@redhat.com>
 
 EXPOSE 8080
 
-ENV NODEJS_VERSION 0.10
+# Add $HOME/node_modules/.bin to the $PATH, allowing user to make npm scripts
+# available on the CLI without using npm's --global installation mode
+ENV NODEJS_VERSION=0.10 \
+    PATH=$HOME/node_modules/.bin/:$PATH
 
 LABEL io.k8s.description="Platform for building and running Node.js 0.10 applications" \
       io.k8s.display-name="Node.js 0.10" \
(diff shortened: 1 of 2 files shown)


0.10/Dockerfile
```diff
@@ -30,6 +30,10 @@ COPY ./s2i/bin/ $STI_SCRIPTS_PATH
 # run and build the applications.
 COPY ./contrib/ /opt/app-root
 
+# Add ./node_modules/.bin to the $PATH, so users won't need to use
+# --global installation (npm i -g) to make npm scripts available on the CLI
+RUN echo "export PATH=\"\$PATH:\$HOME/node_modules/.bin\"" >> /etc/profile.d/custom.sh
```
bparees: the right pattern for appending the path can be seen here:
https://github.com/openshift/sti-python/blob/master/3.3/Dockerfile#L11

(i don't have an opinion as to whether this new item should be first or last in the path, but use the ENV mechanism to add it)

ryanj: Thanks for the feedback - figured there might be a better way.  I've fixed and rebased

bparees:
lgtm, our test infrastructure has some issues at the moment but once we get it fixed and this passes i'll merge it.

bparees:
[test]

ryanj:
Thanks @bparees!

PR #89, opened 2016-04-06
move debug settings into run script

bparees:


bparees:
@Ian-Stewart-Binks @mfojtik i feel like this belongs here instead of in the scl_enable script.

@Ian-Stewart-Binks was there a reason you put it in scl_enable instead?  scl_enable will contain stuff that gets run when a user exec's into a running container.

bparees:
@ryanj fyi

Ian-Stewart-Binks:
@bparees It had something to do with the setting of the environment variables. I recall [this](http://stackoverflow.com/questions/10781824/export-not-working-in-my-shell-script) being relevant.

mfojtik:
@bparees no idea why we have that in SCL enable script, good catch!

mfojtik:
[test]

bparees:
[test]

bparees:
@ryanj @mfojtik i've updated this to fix the env variable testing logic and also print the variable values out when the application starts.  ptal.

users who exec into a running container will not see the same env variables that the node process is running with, since the variables are set by the "run" script based on the value of DEV_MODE, but that doesn't seem like a huge problem to me.

bparees:
@ryanj @mfojtik bump

mfojtik:
LGTM

PR #90, opened 2016-04-11
move ENV defaults to Dockerfile

ryanj:
alternative implementation for #89

### DIFF
--- a/0.10/contrib/etc/scl_enable
+++ b/0.10/contrib/etc/scl_enable
@@ -1,23 +1,3 @@
 # This will make scl collection binaries work out of box.
 unset BASH_ENV PROMPT_COMMAND ENV
 source scl_source enable nodejs010
-
-#Set the debug port to 5858 by default.
-if [ -z "$DEBUG_PORT" ]; then
-  export DEBUG_PORT=5858
-fi
-
-# Set the environment to development by default.
-if [ -z "$DEV_MODE" ]; then
-  export DEV_MODE=false
-fi
-
-# If NODE_ENV is not set by the user, then NODE_ENV is determined by whether
-# the container is run in development mode.
-if [ -z "$NODE_ENV" ]; then
-  if [ "$DEV_MODE" == true ]; then
-    export NODE_ENV=development
-  else
-    export NODE_ENV=production
-  fi
-fi
--- a/0.10/Dockerfile
+++ b/0.10/Dockerfile
@@ -10,7 +10,10 @@ EXPOSE 8080
 # Add $HOME/node_modules/.bin to the $PATH, allowing user to make npm scripts
 # available on the CLI without using npm's --global installation mode
 ENV NODEJS_VERSION=0.10 \
-    PATH=$HOME/node_modules/.bin/:$PATH
+    PATH=$HOME/node_modules/.bin/:$PATH \
+    DEBUG_PORT=5858 \
+    NODE_ENV="production" \
+    DEV_MODE=false
 
 LABEL io.k8s.description="Platform for building and running Node.js 0.10 applications" \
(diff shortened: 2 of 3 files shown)


ryanj:
strange - the test don't seem to be picking up these ENV values

0.10/contrib/etc/scl_enable
```diff
@@ -1,23 +1,3 @@
...
-
-# Set the environment to development by default.
-if [ -z "$DEV_MODE" ]; then
-  export DEV_MODE=false
-fi
-
-# If NODE_ENV is not set by the user, then NODE_ENV is determined by whether
-# the container is run in development mode.
-if [ -z "$NODE_ENV" ]; then
-  if [ "$DEV_MODE" == true ]; then
```
bparees: with this change you're losing the behaviour that setting DEV_MODE controls NODE_ENV.

bparees:
> strange - the tests don't seem to be picking up these ENV values. This solution may need more work.

hmm.. i don't see anything obviously wrong w/ how you're setting them (Aside from the behavior regression..is the test testing the DEV_MODE env variable and failing because you regressed that behavior?)

ryanj:
@bparees, any idea who owns the `DEV_MODE` config option? Is it really needed?

It feels like duplication of variables to me, since `DEV_MODE` = `NODE_ENV != 'production' ? true : false`.

Modifying the config during the container lifecycle seems like an inconsistent way to do config management.  And, it introduces variables that have side-effects on other variables (which is usually easier for app developers to manage via internal config-maps and app-specific logic)

As a NodeJS developer, I'd rather add my repo code into a container image that is clearly pre-configured for "Dev Mode" (a separate "dev mode" image?), OR I'd like to make use of the standard configuration flags for NodeJS environments (`NODE_ENV`).

0.10/contrib/etc/scl_enable
ryanj: Noticed that - I may need to make some changes to the test coverage. 

I'd prefer to drop support for the custom `DEV_MODE` config logic if possible.

bparees:
> @bparees, any idea who owns the DEV_MODE config option? Is it really needed?

it came from @Ian-Stewart-Binks originally, but it's now part of our doc, so we really can't change the behavior w/o breaking backwards compatibility:
https://docs.openshift.org/latest/using_images/s2i_images/nodejs.html#configuration

Ian-Stewart-Binks:
@ryanj My understanding is that a lot of OpenShift tooling is built on top of `$DEV_MODE`, and as such would be trickier to take out.

When I developed this change, I heavily considered simply using `$NODE_ENV` in place of `$DEV_MODE`. It turned out that we required a variable that was distinct from `$NODE_ENV`, that could be set to `true` while `$NODE_ENV` was in [production](https://github.com/openshift/sti-nodejs/blob/master/0.10/test/run#L204). `DEV_MODE = NODE_ENV != 'production' ? true : false` is not correct.

Ian-Stewart-Binks:
> strange - the tests don't seem to be picking up these ENV values. This solution may need more work.

I noted in one of @bparees previous PRs that there was a reason why I chose this implementation over the more initially obvious implementation. If you can find something better, by all means do. In any case, documentation should be updated such that this confusion doesn't happen in the future.

elyscape:
The reason this is happening is a result of the tests' structure. When it checks an environment in the container, [it overwrites some of the environment variables unconditionally, including `NODE_ENV`][get-node-env].

[get-node-env] https://github.com/openshift/sti-nodejs/blob/ba6e4b52995ff5ebd21f189f6b4655e770191a3f/0.10/test/run#L159-L168

PR #93, opened 2016-04-22
init with "npm run -d $NPM_RUN" (w/ default value: "start")

ryanj:
Allow Env control over which script to init with using the `$NPM_RUN` variable

### DIFF
--- a/0.10/README.md
+++ b/0.10/README.md
@@ -85,12 +85,23 @@ Repository organization
 Environment variables
 ---------------------
 
-To set environment variables, you can place them as a key value pair into a `.sti/environment`
-file inside your source code repository.
+Application developers can use the following environment variables to configure the runtime behavior of this image:
+
+NAME        | Description
+------------|-------------
+NODE_ENV    | NodeJS runtime mode (default: "production")
+DEV_MODE    | When set to "true", `nodemon` will be used to automatically reload the server while you work (default: "false"). Setting `DEV_MODE` to "true" will change the `NODE_ENV` default to "development" (if not explicitly set).
+NPM_RUN     | Select an alternate / custom runtime mode, defined in your `package.json` file's [`scripts`](https://docs.npmjs.com/misc/scripts) section (default: npm run "start"). These user-defined run-scripts are unavailable while `DEV_MODE` is in use.
+HTTP_PROXY  | use an npm proxy during assembly
+HTTPS_PROXY | use an npm proxy during assembly
+
+One way to define a set of environment variables is to include them as key value pairs in your repo's `.s2i/environment` file.
 
 Example: DATABASE_USER=sampleUser
 
-Setting the HTTP_PROXY or HTTPS_PROXY environment variable will set the appropriate npm proxy configuration during assembly.
+#### NOTE: Define your own "`DEV_MODE`":
+
+The following `package.json` example includes a `scripts.dev` entry.  You can define your own custom [`NPM_RUN`](https://docs.npmjs.com/cli/run-script) scripts in your application's `package.json` file.
 
 Development Mode
 ---------------------
--- a/0.10/s2i/bin/run
+++ b/0.10/s2i/bin/run
@@ -11,7 +11,7 @@ run_node() {
     exec nodemon --debug="$DEBUG_PORT"
   else
     echo "Launching via npm..."
-    exec npm start -d
+    exec npm run -d $NPM_RUN
   fi
 } 
 
--- a/0.10/Dockerfile
+++ b/0.10/Dockerfile
@@ -10,6 +10,7 @@ EXPOSE 8080
 # Add $HOME/node_modules/.bin to the $PATH, allowing user to make npm scripts
 # available on the CLI without using npm's --global installation mode
 ENV NODEJS_VERSION=0.10 \
+    NPM_RUN=start \
     PATH=$HOME/node_modules/.bin/:$PATH
 
 LABEL io.k8s.description="Platform for building and running Node.js 0.10 applications" \
--- a/0.10/Dockerfile.rhel7
+++ b/0.10/Dockerfile.rhel7
@@ -8,6 +8,7 @@ EXPOSE 8080
 # Add $HOME/node_modules/.bin to the $PATH, allowing user to make npm scripts
 # available on the CLI without using npm's --global installation mode
 ENV NODEJS_VERSION=0.10 \
+    NPM_RUN=start \
     PATH=$HOME/node_modules/.bin/:$PATH
 
 LABEL summary="Platform for building and running Node.js 0.10 applications" \


0.10/s2i/bin/run
```diff
@@ -8,7 +8,7 @@ run_node() {
   if [ "$DEV_MODE" == true ]; then
     exec nodemon --debug="$DEBUG_PORT"
   else
-    exec npm start -d
+    exec npm run $NPM_RUN
```
bparees: should this just be "npm $NPM_RUN" ?  and the default value of NPM_RUN needs to be "start -d"?


also if you're adding a new env variable, please doc it in the README (and ideally in openshift-docs as well, but i can take care of that part).

ryanj: Most people init their server using `npm start`, but `npm run start` is a valid alias. 

I need more info on the `-d` flag. I couldn't find any doc to cover it.  If it's needed, I could change the default config setting from `start` to `start -d`.

This customization would also allow additional init scripts to be added (to cover development modes - or for running `nodemon`, or `gulp`, or `grunt`).

Here are some examples of how that might look:

- http://substack.net/task_automation_with_npm_run
- https://docs.npmjs.com/cli/run-script
- https://docs.npmjs.com/misc/scripts#examples
- https://www.npmjs.com/package/scripty

This should init the example app using nodemon ([similarly to how "DEV_MODE" works](https://github.com/openshift/s2i-nodejs/blob/master/0.10/s2i/bin/run#L9)) by setting `NPM_RUN` to 'dev': https://github.com/openshift/s2ihttps://github.com/openshift/s2i-nodejs-nodejs/blob/master/0.10/test/test-app/package.json#L16 

This would allow developers to define multiple build or dev (or other runtime modes)

ryanj:
np, I'll add some doc for this

0.10/README.md
```diff
@@ -170,3 +170,9 @@ Below is an example _package.json_ file with the _main_ attribute and _start_ sc
 
 #### Note:
 `oc rsync` is only available in versions 3.1+ of OpenShift.
+
+#### Define your own "Dev Mode":
+
+The provided `package.json` example includes a `scripts.dev` entry.  You can define your own custom [`npm run`](https://docs.npmjs.com/cli/run-script) scripts in your application's `package.json` file.
+
+This image allows you to select alternate / custom runtime modes by setting the `$NPM_RUN` environment variable to match one of your available [`scripts`](https://docs.npmjs.com/misc/scripts) entries.
```
bparees: i'd just move this whole section into the `Environment Variables` section, adding an entry for the NPM_RUN env variable.

see this for a better example of how we should be doc'ing env variables the image respects:
https://github.com/openshift/mongodb/tree/master/2.4#environment-variables

ryanj:
@bparees I moved the README updates into the Environment Variables section, and added descriptions for the other dev-facing runtime config settings.

0.10/README.md
```diff
@@ -85,12 +85,23 @@ Repository organization
...
+
+NAME        | Description
+------------|-------------
+NODE_ENV    | NodeJS runtime mode (default: production)
+DEV_MODE    | When set to "true", use `nodemon` to automatically reload the server while you work (default: false)
+NPM_RUN     | Select an alternate / custom runtime mode, defined in your `package.json` file's [`scripts`](https://docs.npmjs.com/misc/scripts) section (default: npm run "start")
+HTTP_PROXY  | set the npm proxy to use during assembly
+HTTPS_PROXY | set the npm proxy to use during assembly
+
+One way to define a set of environment variables is to include them as key value pairs in your repo's `.sti/environment` file.
```
bparees: it's .s2i/environment now (both work but .sti is deprecated so we might as well fix the doc)

0.10/README.md
```diff
@@ -85,12 +85,23 @@ Repository organization
...
 ---------------------
 
-To set environment variables, you can place them as a key value pair into a `.sti/environment`
-file inside your source code repository.
+Application developers can use the following environment variables to configure the runtime behavior of this image:
+
+NAME        | Description
+------------|-------------
+NODE_ENV    | NodeJS runtime mode (default: production)
+DEV_MODE    | When set to "true", use `nodemon` to automatically reload the server while you work (default: false)
```
bparees: setting this to true also sets NODE_ENV to development if it's not explicitly set.  and leaving it as false sets NODE_ENV to production (again, if it's not explicitly set)

bparees:
you learn anything about the mysterious "-d"?

ryanj:
I couldn't find any doc, but it looks like it adds some useful info to the output:

    npm info it worked if it ends with ok
    npm info using npm@2.15.0
    npm info using node@v0.10.44
    npm info prestart http-base@0.1.0
    npm info start http-base@0.1.0

I've updated the PR to include the `-d` flag: https://github.com/openshift/s2i-nodejs/pull/93/files#diff-24840e6960cce9a3cc1e9e312ddbacdd

bparees:
i think the test failure is flaky.
[test]

bparees:
[test]

ryanj:
The tests pass when I run them locally, but the build server keeps reporting: `Yum repo mirrorlist timed out`

bparees:
@ryanj the actual failure is in the npm install:
```
E0504 18:17:56.629851 04048 util.go:91] npm info ok 
npm info it worked if it ends with ok
npm info using npm@2.14.13
npm info using node@v0.10.40
npm info ok 
make: *** [test] Error 1
+ echo 'ERROR: nodejs-rhel7 failed testing.'
```

but i still don't think it's the PRs fault
[test]

mfojtik:
[test]

ryanj:
will this bot listen when I ask for a ... ?
[test]

bparees:
probably not.
[test]

bparees:
ok so i cherry picked and tested this change in what was a working master, and it's definitely breaking:

```
Waiting for container start...
Environment: 
	DEV_MODE=false
	NODE_ENV=production
	DEBUG_PORT=5858
Launching via npm...
Testing the image SCL enable ...
npm info it worked if it ends with ok
npm info using npm@2.14.13
npm info using node@v0.10.40
Lifecycle scripts included in node-echo:
  start
    node server.js

available via `npm run-script`:
  dev
    nodemon --ignore node_modules/ server.js
npm info ok 
ERROR[exec /bin/bash -c node --version] Expected 'v0.10'

```

it looks like the test app is no longer starting up w/ your change?

ryanj:
This fixes the test coverage for me: https://github.com/openshift/s2i-nodejs/pull/93/commits/3a225430cb02af2b23520b6bd98ae8d855e856f1

ryanj:
hmm... `make test` works locally for me, and I don't see a clear error from the build server.

@bparees Let me know if you have any suggestions on how to debug this.

bparees:
i don't think version string was the issue and we don't want to make that check more specific because we don't want to have this start failing for v0.10.41 or something.

the failures appear to be specific to the rhel image build/tests, which you probably can't easily reproduce, i'll see if i can dig into it more later.

ryanj:
ok, I'll remove the attempt to fix the test coverage since it's unrelated to this PR anyway.

bparees:
ah. you need to add the NPM_RUN env variable default to the Dockerfile.rhel7 file.

ryanj:
tests passing!

bparees:
excelllent.

PR #94, opened 2016-04-22
bzip2 is a dependency for installing patternfly, phantomjs

ryanj:
see https://github.com/ryanj/origin-s2i-nodejs/issues/8

### DIFF
--- a/0.10/Dockerfile
+++ b/0.10/Dockerfile
@@ -21,7 +21,7 @@ LABEL io.k8s.description="Platform for building and running Node.js 0.10 applica
       com.redhat.dev-mode.port="DEBUG_PORT:5858"
 
 RUN yum install -y centos-release-scl && \
-    INSTALL_PKGS="nodejs010" && \
+    INSTALL_PKGS="nodejs010 bzip2" && \
     yum install -y --setopt=tsflags=nodocs $INSTALL_PKGS && \
     rpm -V $INSTALL_PKGS && \
     yum clean all -y


ryanj:
Let me know if you'd rather have this added to sti-base.  other language-specific build tooling might need it as well...

ryanj:
The [patternfly](https://www.patternfly.org/) team needs this as well

bparees:
let's leave it here for now, we can always push it down later.

ryanj:
thanks!

elyscape:
@ryanj: Be advised that PhantomJS also needs `freetype` and `fontconfig` in order to run correctly.

elyscape:
Also, this should update `Dockerfile.rhel7` as well.

ryanj:
@elyscape good catch, I'll send an update for `Dockerfile.rhel7`.

It looks like `freetype` and `fontconfig` are available in the centos base image.  

Are they not included in the rhel base image?  Do they need to be added to the rhel Dockerfile as well?

elyscape:
> It looks like `freetype` and `fontconfig` are available in the centos base image.

Interesting. That must have changed, as I definitely had to install them for some of the work I've been doing.

> Are they not included in the rhel base image? Do they need to be added to the rhel Dockerfile as well?

I don't have access to this image, so I couldn't say.

bparees:
@ryanj i already caught the rhel7 miss and included it in my nss pr:
https://github.com/openshift/s2i-nodejs/pull/95/files#diff-debeaecb79f21738a847dd9c4a710bf7R32

PR #95, opened 2016-04-25
add nss support

bparees:
fixes https://github.com/openshift/s2i-base/issues/91

### DIFF
--- a/0.10/contrib/etc/generate_container_user
+++ b/0.10/contrib/etc/generate_container_user
@@ -1,0 +1,17 @@
+# Set current user in nss_wrapper
+USER_ID=$(id -u)
+GROUP_ID=$(id -g)
+
+if [ x"$USER_ID" != x"0" -a x"$USER_ID" != x"1001" ]; then
+
+    NSS_WRAPPER_PASSWD=/opt/app-root/etc/passwd
+    NSS_WRAPPER_GROUP=/etc/group
+
+    cat /etc/passwd | sed -e 's/^default:/builder:/' > $NSS_WRAPPER_PASSWD
+
+    echo "default:x:${USER_ID}:${GROUP_ID}:Default Application User:${HOME}:/sbin/nologin" >> $NSS_WRAPPER_PASSWD
+
+    export NSS_WRAPPER_PASSWD
+    export NSS_WRAPPER_GROUP
+    export LD_PRELOAD=libnss_wrapper.so
+fi
--- a/0.10/Dockerfile
+++ b/0.10/Dockerfile
@@ -21,7 +21,7 @@ LABEL io.k8s.description="Platform for building and running Node.js 0.10 applica
       com.redhat.dev-mode.port="DEBUG_PORT:5858"
 
 RUN yum install -y centos-release-scl && \
-    INSTALL_PKGS="nodejs010 bzip2" && \
+    INSTALL_PKGS="nodejs010 bzip2 nss_wrapper" && \
     yum install -y --setopt=tsflags=nodocs $INSTALL_PKGS && \
     rpm -V $INSTALL_PKGS && \
     yum clean all -y
--- a/0.10/Dockerfile.rhel7
+++ b/0.10/Dockerfile.rhel7
@@ -29,7 +29,7 @@ LABEL com.redhat.component="nodejs010" \
 RUN yum-config-manager --enable rhel-server-rhscl-7-rpms && \
     yum-config-manager --enable rhel-7-server-optional-rpms && \
     yum-config-manager --disable epel >/dev/null || : && \
-    INSTALL_PKGS="nodejs010 nodejs-nodemon" && \
+    INSTALL_PKGS="nodejs010 nodejs-nodemon bzip2 nss_wrapper" && \
     ln -s /usr/lib/node_modules/nodemon/bin/nodemon.js /usr/bin/nodemon && \
     yum install -y --setopt=tsflags=nodocs $INSTALL_PKGS && \
     rpm -V $INSTALL_PKGS && \
--- a/0.10/s2i/bin/run
+++ b/0.10/s2i/bin/run
@@ -2,6 +2,10 @@
 
 set -e
 
+if [ -e "/opt/app-root/etc/generate_container_user" ]; then
+  source /opt/app-root/etc/generate_container_user
+fi
+
 # Runs the nodejs application server. If the container is run in development mode,
 # hot deploy and debugging are enabled.
 run_node() {


bparees:
@ryanj can you pull this down and give it a shot for your issue?

ryanj:
you bet. will let you know how it goes

elyscape:
Why not instead put this in the s2i-base image?

bparees:
@elyscape the base image doesn't have a run/assemble script to add the logic to, and not all the s2i images have, so far, found a need for nss wrapper.

elyscape:
Gotcha. That makes sense.

bparees:
@mfojtik ptal as well.

0.10/contrib/etc/generate_container_user
```diff
@@ -0,0 +1,19 @@
...
+
+    cat /etc/passwd | sed -e 's/^default:/builder:/' > $NSS_WRAPPER_PASSWD
+
+    echo "default:x:${USER_ID}:${GROUP_ID}:Default Application User:${HOME}:/sbin/nologin" >> $NSS_WRAPPER_PASSWD
+
+    export NSS_WRAPPER_PASSWD
+    export NSS_WRAPPER_GROUP
+
+    LD_PRELOAD=libnss_wrapper.so
+    export LD_PRELOAD
```
mfojtik: export LD_PRELOAD=libnss_wrapper.so

0.10/contrib/etc/generate_container_user
```diff
@@ -0,0 +1,19 @@
+# Set current user in nss_wrapper
+USER_ID=$(id -u)
+GROUP_ID=$(id -g)
+
+if [ x"$USER_ID" != x"0" -a x"$USER_ID" != x"1001" ]; then
+
+    NSS_WRAPPER_PASSWD=/opt/app-root/etc/passwd
```
mfojtik: export NSS_WRAPPER_PASSWD

0.10/s2i/bin/assemble
```diff
@@ -7,6 +7,8 @@
 
 set -e
 
+source /opt/app-root/etc/generate_container_user
```
mfojtik: maybe we should check if that script exists? iow. somebody is trying to use assemble outside this image :)

0.10/contrib/etc/generate_container_user
```diff
@@ -0,0 +1,19 @@
...
+    NSS_WRAPPER_GROUP=/etc/group
+
+    cat /etc/passwd | sed -e 's/^default:/builder:/' > $NSS_WRAPPER_PASSWD
```
mfojtik: can you explain the logic here? you replace the "default" user with "builder" and then create new default? also you don't need `cat`...

0.10/contrib/etc/generate_container_user
```diff
@@ -0,0 +1,19 @@
...
+
+    echo "default:x:${USER_ID}:${GROUP_ID}:Default Application User:${HOME}:/sbin/nologin" >> $NSS_WRAPPER_PASSWD
```
mfojtik: are we sure HOME will not change? ;-)

mfojtik:
@bparees couple nits :)

0.10/contrib/etc/generate_container_user
bparees: we don't want to export that until we successfully create the new passwd file.  i know, unlikely to be an issue, but it seems best not to.

0.10/s2i/bin/assemble
bparees: good idea.

0.10/contrib/etc/generate_container_user
bparees: user 1001 exists in our /etc/passwd file because sti-base provided it, but really that user is the "builder" user (it's what assemble runs as).  

Since we need to create a new default user to represent the random id this container is running as, we first need to rename the old "default" user to something else so there is no conflict (we could also just filter it out entirely) and renaming it to builder seems logical.

this actually came from @GrahamDumpleton's suggested fix to the python image.

feline removed.

0.10/contrib/etc/generate_container_user
bparees: change to what?  this is the same way we do this everywhere else when we construct a user for nss wrapper.

bparees:
@mfojtik updated/responded

0.10/contrib/etc/generate_container_user
mfojtik: @bparees thanks! it makes sense... the assemble will also get the UID mapping right I assume.

0.10/contrib/etc/generate_container_user
mfojtik: nvmd :-)

mfojtik:
LGTM (merge when tests pass :-)

0.10/contrib/etc/generate_container_user
bparees: the assemble container will run as uid 1001 so all this logic will be skipped (nss will not be set up) and since the /etc/passwd contains a valid entry for uid 1001, things will be fine (whoami will reported "default" since that's the entry in /etc/passwd)

bparees:
@ryanj did you get a chance to try this?

ryanj:
@bparees @mfojtik 
I gave this a try, and the good news is that `npm install bower` is finishing cleanly now.

The bad news, is that I'm having trouble reproducing [the original issue that this change was meant to address](https://github.com/openshift/s2i-base/issues/91)  :(((

Not sure if we should merge it, or put this branch on ice for now, until we can find a solid reproduction case for the original issue.  

Not sure if this is significant, but `whoami` is currently returning:

    whoami: cannot find name for user ID 1000110000

Thanks for looking into this. I'll report back if I see the issue come up again.

bparees:
@ryanj under what circumstances did you run the "whoami"?  rsh'd into a running pod?  if so that's to be expected since the rsh session doesn't have the nss environment setup.

GrahamDumpleton:
FWIW. For Python issue about adding nss to shell env init stuff is https://github.com/openshift/s2i-python/issues/96.

bparees:
i've removed the nss setup for the assemble path since assemble runs as a "real" user (uid 1001 which appears in /etc/passwd).

@ryanj any updates on whether you even still have this issue or how to recreate it?

ryanj:
I think the issue originally happened in the online dev-preview environment.  I haven't seen it come up since.

bparees:
hm.  well in the spirit of "it can't hurt and it'll align it w/ other images and maybe save someone some trouble down the road" i'm going to go ahead and merge it anyway.

ryanj:
sounds good - I'll make sure to add the same thing in ryanj/origin-s2i-nodejs

PR #96, opened 2016-05-06
Nodejs 4.4 repository

kasicka:
New repository containing files for creating Nodejs 4.4 s2i images from RHSCL 2.2 beta release


4.4/Dockerfile
```diff
@@ -0,0 +1,44 @@
...
+      io.k8s.display-name="Node.js 4.4.2" \
+      io.openshift.expose-services="8080:http" \
+      io.openshift.tags="builder,nodejs,nodejs4" \
+      com.redhat.dev-mode="DEV_MODE:false" \
+      com.redhat.deployments-dir="/opt/app-root/src" \
+      com.redhat.dev-mode.port="DEBUG_PORT:5858"
+
+RUN yum install -y centos-release-scl-rh && \
+    yum-config-manager --enable centos-sclo-rh-testing && \
+    INSTALL_PKGS="rh-nodejs4 bzip2" && \
```
mfojtik: i do believe the bzip2 is already installed in the base image.

4.4/README.md
```diff
@@ -0,0 +1,172 @@
...
+This repository contains the source for building various versions of
+the Node.JS application as a reproducible Docker image using
+[source-to-image](https://github.com/openshift/source-to-image).
+Users can choose between RHEL and CentOS based builder images.
+The resulting image can be run using [Docker](http://docker.io).
+
+
+Usage
+---------------------
+To build a simple [nodejs-sample-app](https://github.com/openshift/sti-nodejs/tree/master/4.3/test/test-app) application
```
mfojtik: update link?

4.4/s2i/bin/assemble
```diff
@@ -0,0 +1,48 @@
...
+# The official nodejs:4.4-onbuild already run npm install and use different
+# application folder.
+[ -d "/usr/src/app" ] && exit 0
+
+set -e
+
+# FIXME: Linking of global modules is disabled for now as it causes npm failures
+#        under RHEL7
+# Global modules good to have
+# npmgl=$(grep "^\s*[^#\s]" ../etc/npm_global_module_list | sort -u)
```
mfojtik: is this still broken? (you copying the list to the image but not using it). can you try to uncomment this and check if the tests passing?

4.4/Dockerfile
kasicka: There's bzip2-libs package only

4.4/contrib/etc/scl_enable
```diff
@@ -0,0 +1,23 @@
...
+
+#Set the debug port to 5858 by default.
+if [ -z "$DEBUG_PORT" ]; then
+  export DEBUG_PORT=5858
+fi
+
+# Set the environment to development by default.
+if [ -z "$DEV_MODE" ]; then
+  export DEV_MODE=false
+fi
```
ryanj: I think this code was relocated in https://github.com/openshift/s2i-nodejs/pull/89

4.4/s2i/bin/assemble
kasicka: Failing so far..

bparees:
[test]

bparees:
[test]

bparees:
[test]

bparees:
I looked into this tonight, a couple items:
1) the centos docker image is now building fine
2) the rhel image cannot be built/tested.  
  * When using the rhscl base image, the primary issue is that the rhscl base image does not include subcsription certs, so we can't install anything.
  * When using the openshift base image, the issue is that the rhel-server-rhscl-beta-7-rpms repo isn't available.  If someone can give me a .repo file for that repo, i can try to make it work.

but the implication for the scl images in general is that, if we're going to keep using the openshift build/test infrastructure to validate these images (not build the published images), they need to be based on the "openshift/base-rhel7" image because within our infrastructure that's the only image that'll have proper subscription credentials.

rhcarvalho:
bump

Now that SCL 2.2 is GA we don't need to use the beta repo.

hhorak:
@kasicka The last two commits don't make it better..
SCLs for CentOS are still not in mirrors, so `centos-sclo-rh-testing` repo is still the only thing how to get to the packages on CentOS.
On RHEL, what we need actually is `s/rhel-server-rhscl-beta-7-rpms/rhel-server-rhscl-7-rpms/`, not removing the line entirely.

4.4/Dockerfile.rhel7
```diff
@@ -0,0 +1,52 @@
+#FROM openshift/base-rhel7
+FROM registry.access.redhat.com/rhscl/s2i-base-rhel7
```
bparees: after some discussion with @hhorak, you can make this rhscl/s2i-base-rhel7

bparees:
i've pushed changes such that our test processes will now ignore the rhel versions of the SCL images so we don't have repo issues.  please rebase this PR and we can see if it'll get through testing (I assume you can also remove all the beta related repos now since the rpms are in the GA repo)

bparees:
[test]

hhorak:
[test]

bparees:
[test]

bparees:
this still needs a rebase before it can be properly tested, to pick up the changes that will skip rhel testing.

bparees:
also it's still not passing tests even w/ the rebase (I did a rebase my own PR to test it):
https://github.com/sclorg/s2i-nodejs-container/pull/104

sdodson:
If you're building on an entitled host the image build should inherit
access to all repos that the host is entitled to.

bparees:
> If you're building on an entitled host the image build should inherit access to all repos that the host is entitled to.

yeah, but we aren't.  the ec2 instances we spin up to build the rhel images for testing don't have a subscription.  we work around that in a hacky way for images based on "base-rhel7" but images based on "scl-base-rhel7" can't take advantage of the hack, so in the end we've just disabled testing the SCL rhel-based images on our ec2 instances.  But this PR needs to be rebased to pick up that change.

bparees:
that rebase does not look to have been done correctly.

bparees:
@kasicka bump

4.4/Dockerfile.rhel7
```diff
@@ -0,0 +1,58 @@
...
+      name="openshift3/nodejs-4-rhel7" \
+      version="4.4" \
+      release="1" \
+      architecture="x86_64"
+
+RUN yum-config-manager --enable rhel-server-rhscl-7-rpms && \
+    # yum-config-manager --enable rhel-server-rhscl-beta-7-rpms && \
+    yum-config-manager --enable rhel-7-server-optional-rpms && \
+    yum-config-manager --disable epel >/dev/null || : && \
+    INSTALL_PKGS="rh-nodejs4 rh-nodejs4-nodejs-nodemon" && \
```
bparees: why isn't this installing npm?

bparees: i guess the other packages pull it in as a dep.

ryanj:
Is the NodeJS release number the only change in this new folder of files?  Duplicating all of this code may introduce a maintenance burden.  

I know this doesn't match the current conventions of 1 folder per runtime version - but here is an example where I reuse code rather than duplicating it, generating all releases from a single folder: https://github.com/ryanj/origin-s2i-nodejs/

Let me know if anyone would be interested in a PR that abstracts the NodeJS release version to offer a generalized solution with fewer folders of code to maintain.

hhorak:
Closing this PR, since it was already provided via #109.
@ryanj See https://github.com/sclorg/mysql-container/pull/143, there is a PoC of de-duplication, that we might use.

PR #99, opened 2016-05-15
Install nodemon from SCL

hhorak:
No need to create symlink either, since nodemon is installed into
/opt/rh/nodejs010/root/usr/bin/nodemon and nodejs010 is enabled.
We also have nodemon in centos already.


bparees:
thanks @hhorak

PR #102, opened 2016-05-24
update usage info, store globals locally

ryanj:
Users shouldn't need to install tools globally - but I'm sure they'll find a way to do it regardless.

When they do, we should store these globals locally, just like every other part of our application's build context.

### DIFF
--- a/0.10/Dockerfile
+++ b/0.10/Dockerfile
@@ -7,11 +7,13 @@ MAINTAINER SoftwareCollections.org <sclorg@redhat.com>
 
 EXPOSE 8080
 
-# Add $HOME/node_modules/.bin to the $PATH, allowing user to make npm scripts
-# available on the CLI without using npm's --global installation mode
-ENV NODEJS_VERSION=0.10 \
-    NPM_RUN=start \
-    PATH=$HOME/node_modules/.bin/:$PATH
+# This image will be initialized with "npm run $NPM_RUN"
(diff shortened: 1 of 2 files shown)


0.10/Dockerfile
```diff
@@ -7,11 +7,13 @@ MAINTAINER SoftwareCollections.org <sclorg@redhat.com>
...
-ENV NODEJS_VERSION=0.10 \
-    NPM_RUN=start \
-    PATH=$HOME/node_modules/.bin/:$PATH
+# This image will be initialized with "npm run $NPM_RUN"
+# See https://docs.npmjs.com/misc/scripts, and your repo's package.json
+# file for possible values of NPM_RUN
+ENV NPM_RUN=start \
+    NODEJS_VERSION=0.10 \
+    NPM_CONFIG_PREFIX=$HOME/.npm-global \
+    PATH=$HOME/.npm-global/bin/:$HOME/node_modules/.bin/:$PATH
```
bparees: should global really be the first item in the path?

bparees:
what part of this is "update usage info" ?

0.10/Dockerfile
ryanj: good call - I switched the order so that the package.json dependencies and release version info will get highest priority.

ryanj:
The "usage info" is line notes, directly in the Dockerfile (around line 10)

bparees:
ah, thought it was referring to changes to the usage script.
lgtm.

PR #104, opened 2016-06-15
[DO_NOT_MERGE] test nodejs4

bparees:
testing @kasicka's nodejs4 PR after rebasing it to see if the v4 rhel image is still getting built/tested (it should not be)

### DIFF
--- a/0.10/README.md
+++ b/0.10/README.md
@@ -47,7 +47,7 @@ Repository organization
 
     * **`s2i/bin/`**
 
-        This folder contains scripts that are run by [STI](https://github.com/openshift/source-to-image):
+        This folder contains scripts that are run by [S2I](https://github.com/openshift/source-to-image):
 
         *   **assemble**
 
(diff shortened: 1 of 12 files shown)


bparees:
[test]

bparees:
Seems to still be some issues in the nodejs4 image, not passing tests:
I0615 18:07:49.437621 19441 clone.go:33] Downloading "/tmp/image_test.ODLVgnq/sti-nodejs/4.4/test/test-app" ...
I0615 18:07:49.455274 19441 install.go:251] Using "assemble" installed from "image:///usr/libexec/s2i/assemble"
I0615 18:07:49.455313 19441 install.go:251] Using "run" installed from "image:///usr/libexec/s2i/run"
I0615 18:07:49.455333 19441 install.go:251] Using "save-artifacts" installed from "image:///usr/libexec/s2i/save-artifacts"
E0615 18:07:50.066617 19441 util.go:96] cp: cannot stat '/tmp/src/.': No such file or directory
E0615 18:07:50.395939 19441 main.go:352] An error occurred: non-zero (13) exit code from centos/nodejs-44-centos7-candidate
make: *** [test] Error 1

PR #105, opened 2016-07-01
Add NPM_MIRROR env var to NodeJS images

dinhxuanvu:
The NPM_MIRROR is introduced to allow users to use alternative custom
NPM registry mirror during the build process. The NPM packages will be
downloaded from the custom link instead of the default 'registry.npmjs.org'.

Signed-off-by: Vu Dinh <vdinh@redhat.com>


bparees:
lgtm.

bparees:
(waiting for test success)

PR #108, opened 2016-07-06
disable gpg signed commits (if enabled) when generating test artifacts

ryanj:
I have 'signed commits' enabled on my dev machine, but I don't want `make test` to prompt me for my signing key while running tests


bparees:
lgtm

PR #110, opened 2016-07-08
Use just 4 as compatible part of the version

hhorak:
Currently we use 4.4 in Dockerfile and also in name of resulting centos image, while the SCL name only uses 4 as a version and also RHEL image is available under nodejs-4-rhel7 name. Thus, I think we should use the same version scheme for centos as well, i.e. centos/nodejs-4-centos7 and not centos/nodejs-44-centos7. I also think it might happen that the image will eventually include nodejs 4.5.x. But even when not, I don't expect we'd want to maintain different versions of nodejs 4 image, e.g. for 4.5 and 4.4.
This PR should fix the sources, but I'm not sure what needs to happen in the `push` scripts.
@kasicka, please confirm this is what we want, before merging.


hhorak:
[test]

ryanj:
This goes against convention a bit, but it might allow you to avoid some of the name/version maintenance complications if you stick with a single image on DockerHub at `centos/nodejs-centos7` or `centos7-s2i-nodejs`.  Individual runtimes could be published as tags: `centos/centos7-s2i-nodejs:0.10`, `centos/centos7-s2i-nodejs:4.4`, `centos/centos7-s2i-nodejs:4`, `centos/centos7-s2i-nodejs:lts`, etc

bparees:
i would prefer this as well

bparees:
by this i mean the -4- instead of -44- naming.

as far as i know the push for centos images is still being handled by us (openshift) so i'll make sure that works once this merges.

as for the duplication, it is a bit painful at times, but we've managed it in all our existing repos so far and i kinda prefer it over handling the inevitable special casing as different versions diverge in different ways and addressing bugs in scripts that enable common file sharing.  But at this point it's up to the SCL team.

bparees:
@hhorak @kasicka bump, i'm waiting for this to go in to push the nodejs4 centos image to dockerhub and update the openshift imagestreams to include it.

kasicka:
+1 from me

PR #114, opened 2016-08-10
Add bzip2 to Dockerfiles

kasicka:
Customer can't use the image, because the bzip2 package is missing.
https://bugzilla.redhat.com/show_bug.cgi?id=1364546


bparees:
should we consider adding this to the base image instead?

bparees:
@hhorak

hhorak:
yum tells me it's 77kB, so I'd also be in favor to have it in s2i base image. @kasicka, would you mind creating a PR for s2i base instead?

PR #118, opened 2016-09-12
Configurable npm loglevel. Allow npm --loglevel to be set through environment variable

PhilipGough:
Motivation for this change is to allow the user to specify the verbosity level of the out put from the npm install -- loglevel flag. The proposed change here is to set an environment variable which would be checked before the build and default to npms own default setting of 'warn' if not set

### DIFF
--- a/0.10/README.md
+++ b/0.10/README.md
@@ -92,9 +92,9 @@ NAME        | Description
 NODE_ENV    | NodeJS runtime mode (default: "production")
 DEV_MODE    | When set to "true", `nodemon` will be used to automatically reload the server while you work (default: "false"). Setting `DEV_MODE` to "true" will change the `NODE_ENV` default to "development" (if not explicitly set).
 NPM_RUN     | Select an alternate / custom runtime mode, defined in your `package.json` file's [`scripts`](https://docs.npmjs.com/misc/scripts) section (default: npm run "start"). These user-defined run-scripts are unavailable while `DEV_MODE` is in use.
-HTTP_PROXY  | use an npm proxy during assembly
-HTTPS_PROXY | use an npm proxy during assembly
-NPM_MIRROR  | use a custom NPM registry mirror to download packages during the build process
+HTTP_PROXY  | Use an npm proxy during assembly
+HTTPS_PROXY | Use an npm proxy during assembly
+NPM_MIRROR  | Use a custom NPM registry mirror to download packages during the build process
 
 One way to define a set of environment variables is to include them as key value pairs in your repo's `.s2i/environment` file.
 
@@ -103,6 +103,9 @@ Example: DATABASE_USER=sampleUser
 #### NOTE: Define your own "`DEV_MODE`":
 
 The following `package.json` example includes a `scripts.dev` entry.  You can define your own custom [`NPM_RUN`](https://docs.npmjs.com/cli/run-script) scripts in your application's `package.json` file.
+
+#### Note: Setting logging output verbosity
+To alter the level of logs output during an `npm install` the npm_config_loglevel environment variable can be set. See [npm-config](https://docs.npmjs.com/misc/config).
 
 Development Mode
 ---------------------
--- a/0.10/s2i/bin/assemble
+++ b/0.10/s2i/bin/assemble
@@ -22,6 +22,9 @@ shopt -s dotglob
 echo "---> Installing application source ..."
 mv /tmp/src/* ./
 
+echo "--> Setting npm loglevel to ${NPM_LOG_LEVEL:-warn}"
+npm config set loglevel ${NPM_LOG_LEVEL:-warn}
+
 if [ ! -z $HTTP_PROXY ]; then
 	echo "---> Setting npm http proxy to $HTTP_PROXY"
 	npm config set proxy $HTTP_PROXY
@@ -48,7 +51,7 @@ if [ -n "$NPM_MIRROR" ]; then
 fi
 
 echo "---> Building your Node application from source"
-npm install -d
+npm install
 
 # Fix source directory permissions
 fix-permissions ./
--- a/0.10/s2i/bin/assemble
+++ b/0.10/s2i/bin/assemble
@@ -22,9 +22,6 @@ shopt -s dotglob
 echo "---> Installing application source ..."
 mv /tmp/src/* ./
 
-echo "--> Setting npm loglevel to ${NPM_LOG_LEVEL:-warn}"
-npm config set loglevel ${NPM_LOG_LEVEL:-warn}
-
 if [ ! -z $HTTP_PROXY ]; then
 	echo "---> Setting npm http proxy to $HTTP_PROXY"
 	npm config set proxy $HTTP_PROXY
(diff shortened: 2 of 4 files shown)


0.10/s2i/bin/assemble
```diff
@@ -47,8 +47,20 @@ if [ -n "$NPM_MIRROR" ]; then
 	npm config set registry $NPM_MIRROR
 fi
 
+if [ ! -z $NPM_LOG_LEVEL && $NPM_LOG_LEVEL == "info" ]; then
+	echo "--> ${NPM_LOG_LEVEL} set. Setting npm logging level to 'info'"
+	npm config set loglevel info
+
```
rhcarvalho: inconsistent new line?

0.10/s2i/bin/assemble
```diff
@@ -47,8 +47,20 @@ if [ -n "$NPM_MIRROR" ]; then
 	npm config set registry $NPM_MIRROR
 fi
 
+if [ ! -z $NPM_LOG_LEVEL && $NPM_LOG_LEVEL == "info" ]; then
+	echo "--> ${NPM_LOG_LEVEL} set. Setting npm logging level to 'info'"
```
rhcarvalho: This message will look strange after variable expansion:

```
--> silent set. Setting npm logging level to 'info'
```

0.10/s2i/bin/assemble
```diff
@@ -47,8 +47,20 @@ if [ -n "$NPM_MIRROR" ]; then
...
+	npm config set loglevel info
+
+elif [ ! -z $NPM_LOG_LEVEL && $NPM_LOG_LEVEL == "silent" ]; then
+	echo "--> ${NPM_LOG_LEVEL} set. Setting npm logging level to 'silent'"
+	npm config set loglevel silent
+else
+	echo "--> No NPM_LOG_LEVEL environment variable set. Setting npm logging level to default 'warn'"
+	npm config set loglevel warn
+fi
```
rhcarvalho: I think you can replace all this if..else with a single block:

```bash
echo "--> Setting NPM loglevel to ${NPM_LOG_LEVEL:-warn}"
npm config set loglevel ${NPM_LOG_LEVEL:-warn}
```

?

And since this affects calls to the `npm` command, I think this should probably be done earlier in the top of the script?

0.10/README.md
```diff
@@ -87,14 +87,15 @@ Environment variables
...
-HTTPS_PROXY | use an npm proxy during assembly
-NPM_MIRROR  | use a custom NPM registry mirror to download packages during the build process
+NAME          | Description
+--------------|-------------
+NODE_ENV      | NodeJS runtime mode (default: "production")
+DEV_MODE      | When set to "true", `nodemon` will be used to automatically reload the server while you work (default: "false"). Setting `DEV_MODE` to "true" will change the `NODE_ENV` default to "development" (if not explicitly set).
+NPM_RUN       | Select an alternate / custom runtime mode, defined in your `package.json` file's [`scripts`](https://docs.npmjs.com/misc/scripts) section (default: npm run "start"). These user-defined run-scripts are unavailable while `DEV_MODE` is in use.
+HTTP_PROXY    | use an npm proxy during assembly
+HTTPS_PROXY   | use an npm proxy during assembly
+NPM_MIRROR    | use a custom NPM registry mirror to download packages during the build process
```
rhcarvalho: Since you are touching this already, could you make the descriptions consistent by starting all of them with capital case? s/use/Use above and s/specify/Specify below.

PhilipGough:
@rhcarvalho Thanks for review. Will update based on comments

0.10/README.md
```diff
@@ -87,14 +87,15 @@ Environment variables
...
+NPM_LOG_LEVEL | specify the logging output for npm. Options ("info","warn","silent"). Default to warn.
```
rhcarvalho: The options shouldn't be this restrictive, we actually allow all values that npm understands `npm help 7 config` says silent, warn, info, verbose and silly.
I would document this by simply saying "any value that npm understands, e.g., silent, warn, info, verbose and silly"

0.10/s2i/bin/assemble
bparees: still need an if check to confirm that NPM_LOG_LEVEL is set to one of "info", "warn", or "silent" (or is empty)

but yes that can be a single if.

PhilipGough: @bparees I think this makes sense actually because we can just assign the default value to warn if its not set

PhilipGough:
@rhcarvalho @bparees Updated based on review. Thanks

rhcarvalho:
Would be nice to see how this changes the output of an example build like https://github.com/openshift/nodejs-ex/. I expect to see less garbage by default (due to the removal of the `-d` flag).

I'm still in doubt if it is a good idea to apply the log level change globally, or if we want to set it just for the install part (we could take an env var to pass arbitrary args to `npm install`?!)

Another thing I read about in the npm help is that it is possible to configure independent loglevels per package, and with the env var solution here we don't get that flexibility. We probably don't need it now either, and if we one day allow passing in a custom config file, the config file could take precedence.

0.10/s2i/bin/assemble
```diff
@@ -22,6 +22,9 @@ shopt -s dotglob
 echo "---> Installing application source ..."
 mv /tmp/src/* ./
 
+echo "--> Setting npm loglevel to ${NPM_LOG_LEVEL:-warn}"
+npm config set loglevel ${NPM_LOG_LEVEL:-warn}
```
rhcarvalho: I know I suggested the above, but on second thought I think it is better to follow the pattern set by the other env vars.

```bash
 if [ ! -z $NPM_LOG_LEVEL ]; then
  	echo "---> Setting npm loglevel to $NPM_LOG_LEVEL"
  	npm config set loglevel $NPM_LOG_LEVEL
 fi
```

The advantage here is that if NPM_LOG_LEVEL is not set we do nothing, instead of changing the config. That's preferable because if there is any way to provide custom configuration (.npmrc?) we would not interfere with values set there.

PhilipGough:
Yes I thought about this option also @rhcarvalho . Do you think that would be better option?

PhilipGough:
The precedence is cli arg > env var > .npmrc so my thinking is we should do something like this

```bash
case $NPM_LOG_LEVEL in
    silent|error|warn|http|info|verbose|silly) npm install {$NMP_LOG_LEVEL};;
    *)             npm install;;
esac
```
What do you think? I like this solution because it defaults to "warn" if the env var is not set. It will also allow a user to set this value in the .npmrc file should they not want to set an env var

PhilipGough:
Have updated PR based on suggestion above. Let me know what you guys think. Thanks

0.10/s2i/bin/assemble
```diff
@@ -48,7 +48,10 @@ if [ -n "$NPM_MIRROR" ]; then
 fi
 
 echo "---> Building your Node application from source"
-npm install -d
+case $NPM_LOG_LEVEL in
+    silent|error|warn|http|info|verbose|silly) npm install --loglevel {$NMP_LOG_LEVEL};;
+    *)             npm install;;
```
bparees: so we're changing the default logging level from "info" to "warn"?

PhilipGough: Well yeah because the default npm value is warning. The info is quiet verbose and is what is currently present with the -d flag

rhcarvalho: > so we're changing the default logging level from "info" to "warn"?

Yes.

AFAIK the initial motivation for this is that `npm install -d` was printing way more output than desired (and than the default).

PhilipGough:
@rhcarvalho  Updated based on conversation over IRC

rhcarvalho:
LGTM

bparees:
[test]

rhcarvalho:
FTR, this is a similar PR we had to add a new environment variable in the Python image:
https://github.com/sclorg/s2i-python-container/pull/124

Just like here, it turned out that `pip` (like `npm`) already supported passing arbitrary configuration via environment variables following a certain pattern.

PR #127, opened 2016-11-14
Enable OSE repo for nss_wrapper.

omron93:
In RHEL7 Dockerfile nss_wrapper is installed but OSE repository is not enabled. So installation fails.

@hhorak @bparees Please take a look and merge.


hhorak:
LGTM.

PR #128, opened 2016-11-18
deprecate 0.10

bparees:


bparees:
@hhorak fyi.

hhorak:
Sounds good to me :)

PR #141, opened 2017-05-15
Add NodeJS 6 Dockerfile

hhorak:


omron93:
@hhorak It seems that new packages are not yet ready in CentOS.

hhorak:
The packages are not yet available in the CentOS testing/release repos. It's requested here: https://bugs.centos.org/view.php?id=13261

hhorak:
I've removed v6 from Makefile temporarily so we can merge this. The differences from v4 are available at https://gist.github.com/hhorak/7bfca651736d372cd42642c8b627f4a1.

hhorak:
Missing testing for v6 is tracked in #142.

6/test/run
```diff
@@ -0,0 +1,246 @@
...
+  # should remove this restriction in the future when a file:// is used.
+  pushd ${test_dir}/test-app >/dev/null
+  git init
+  git config user.email "build@localhost" && git config user.name "builder"
+  git add -A && git commit --no-gpg-sign -m "Sample commit"
+  popd >/dev/null
+}
+
+run_test_application() {
+  docker run --user=100001 --rm --cidfile=${cid_file} -p ${test_port} $1 ${IMAGE_NAME}-testapp
```
omron93: Please do not publish container's port to the host - application in container is accessed by container IP, so this is not necessary for tests. And running tests with this can cause conflicts with other containers in host system using same port...

omron93:
Otherwise it looks good to me. Thanks.

4/Dockerfile.rhel7
```diff
@@ -1,4 +1,4 @@
-FROM rhscl/s2i-base-rhel7
+FROM rhscl/s2i-base-rhel7:1
```
omron93: @hhorak What changed please? Should centos based image tagged too?

PR #143, opened 2017-05-24
Move `contrib` to `root/opt/app-root` and update common

torsava:
common-container-scripts new automatically compile the README.md file into a
help.1 and place it into the root subdirectory to be copied to the image during
build.


omron93:
Thanks.

LGTM + tests are passing. So merging.

PR #150, opened 2017-09-20
Remove nodejs6 packages in nodejs4 image

hhorak:
NodeJS 6 gets installed in s2i-base, but it doesn't make sense in rh-nodejs4 image, so this PR removes it.

However, even if this makes the image's content smaller, in the end users might need to download more bytes (me not sure how overlayfs works in case when we install something in one layer and then remove it in another)..

Anyway, I'm personally not entirely convinced whether the space issue is more important than having some packages installed and not used, since there is always some risk of either using wrong packages by mistake, or that a potential attacker will have more tools available.. So, any opinion about this is welcome.

### DIFF
--- a/4/Dockerfile.rhel7
+++ b/4/Dockerfile.rhel7
@@ -44,7 +44,7 @@ RUN yum repolist > /dev/null && \
     yum-config-manager --enable rhel-server-rhscl-7-rpms && \
     yum-config-manager --enable rhel-7-server-rpms && \
     yum-config-manager --enable rhel-7-server-optional-rpms && \
-    yum remove rh-nodejs6\* && \
+    yum remove -y rh-nodejs6\* && \
     INSTALL_PKGS="rh-nodejs4 rh-nodejs4-npm rh-nodejs4-nodejs-nodemon nss_wrapper" && \
     ln -s /usr/lib/node_modules/nodemon/bin/nodemon.js /usr/bin/nodemon && \
     yum install -y --setopt=tsflags=nodocs $INSTALL_PKGS && \
--- a/4/Dockerfile.rhel7
+++ b/4/Dockerfile.rhel7
@@ -44,6 +44,7 @@ RUN yum repolist > /dev/null && \
     yum-config-manager --enable rhel-server-rhscl-7-rpms && \
     yum-config-manager --enable rhel-7-server-rpms && \
     yum-config-manager --enable rhel-7-server-optional-rpms && \
+    yum remove rh-nodejs6\* && \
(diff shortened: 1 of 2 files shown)


hhorak:
@praiskup @pkubatrh @bparees @kasicka Any opinions?

praiskup:
The removal should help at least to squashed (RHEL) images.  But I don't think it is worth removal tbh, since that's another yum statement to take care of.

> there is always some risk of either using wrong packages by mistake,

Maybe we could invent some declarative set of components which are there "intentionally".

>  that a potential attacker will have more tools available

I personally view this as non-issue,  since I'm not security guy.  For us it means some more respins in the worst case if there were some security issues.

Anyway, the interesting question is IMO why s2i-base needs that package, and how to remove it from there.

praiskup:
> The removal should help at least to squashed (RHEL) images.

False, it didn't help. Maybe we want to remove the `*-environment` package which all the collection packages depend on?

pkubatrh:
>However, even if this makes the image's content smaller, in the end users might need to download more bytes

I expected to find that this removal of rh-nodejs6-* packages adds a few bytes into the layer but (at least on my box) it actually makes it smaller by 2MB. Squashed (centos) image is smaller by ~53MB as expected.

hhorak:
> Anyway, the interesting question is IMO why s2i-base needs that package, and how to remove it from there.

That's actually a feature, s2i-base includes nodejs6-npm intentionally, because for many images npm is useful to install static JavaScript content. It just doesn't make sense to have it in rh-nodejs4 image.

4/Dockerfile.rhel7
```diff
@@ -44,6 +44,7 @@ RUN yum repolist > /dev/null && \
     yum-config-manager --enable rhel-server-rhscl-7-rpms && \
     yum-config-manager --enable rhel-7-server-rpms && \
     yum-config-manager --enable rhel-7-server-optional-rpms && \
+    yum remove rh-nodejs6\* && \
```
pkubatrh: Missing `-y`

praiskup:
> Squashed (centos) image is smaller by ~53MB as expected.

Ok, I must be doing something wrong then. :) But my images are of the same size.

pkubatrh:
>Ok, I must be doing something wrong then. :) But my images are of the same size.

Actually I have done something wrong as I have squashed the whole image and not only the layers added on top of s2i-base.
No major size decrease when using `-f docker.io/centos/s2i-base-centos7`

bparees:
no strong opinions either way but i guess i lean slightly towards "sure, if it's not needed (and we can't remove it from the common base), let's remove it."

hhorak:
[test]

4/Dockerfile.rhel7
praiskup: @hhorak still missing, I think

PR #155, opened 2017-10-04
Add symlink to test-lib-openshift.sh

pvalena:
[Changes Info](https://github.com/sclorg/s2i-python-container/pull/218#issuecomment-335276731)

### DIFF
--- a/test/test-lib-openshift.sh
+++ b/test/test-lib-openshift.sh
@@ -1,0 +1,1 @@
+../common/test-lib-openshift.sh


pvalena:
PTAL.

pkubatrh:
Are these going to be modified for nodejs8 now that it is out?

hhorak:
These templates should be part of the nodejs-ex example application.

pvalena:
> These templates should be part of the nodejs-ex example application.

Correct. I'll move them to nodejs-ex repository instead. My bad.


> Are these going to be modified for nodejs8 now that it is out?

Anyway, this needs to be done in 3 steps:

- import already existing template (we're upstream now) 
- add simple tests for templates and ensure they pass
- update templates

pkubatrh:
Is there any reason to keep this open?

pvalena:
This can be actually merged now :)

hhorak:
> Is there any reason to keep this open?

I don't think so, #172 delivers what was supposed to be delivered by this PR. Closing.

PR #163, opened 2017-10-24
Sync downstream changes

pkubatrh:


pkubatrh:
Modified the RUN command so that it does not introduce another layer.

pkubatrh:
Modified the RUN command so that it does not introduce another layer.

PR #168, opened 2017-11-20
tests: replace removed s2i flag

pkubatrh:
Similarily to: sclorg/s2i-python-container#229


omron93:
Thanks for the fix.

Tests are passing. Meging.

PR #170, opened 2017-11-21
Do not log proxy user and password

kasicka:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1504545

### DIFF
--- a/test/run
+++ b/test/run
@@ -45,6 +45,10 @@ run_s2i_build() {
   s2i build ${s2i_args} file://${test_dir}/test-app ${IMAGE_NAME} ${IMAGE_NAME}-testapp
 }
 
+run_s2i_build_proxy() {
+    s2i build ${s2i_args} -e HTTP_PROXY=$1 -e http_proxy=$1 -e HTTPS_PROXY=$2 -e https_proxy=$2 file://${test_dir}/test-hw ${IMAGE_NAME} ${IMAGE_NAME}-testhw
+}
+
 prepare() {
   if ! image_exists ${IMAGE_NAME}; then
     echo "ERROR: The image ${IMAGE_NAME} must exist before this script is executed."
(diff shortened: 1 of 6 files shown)


pkubatrh:
@kasicka Can we get a test-case for this as well? Seems like an easy enough feature to test.

pkubatrh:
[test]

test/run
```diff
@@ -45,34 +45,50 @@ run_s2i_build() {
...
+
 prepare() {
   if ! image_exists ${IMAGE_NAME}; then
     echo "ERROR: The image ${IMAGE_NAME} must exist before this script is executed."
     exit 1
   fi
   # TODO: STI build require the application is a valid 'GIT' repository, we
   # should remove this restriction in the future when a file:// is used.
-  pushd ${test_dir}/test-app >/dev/null
+  if [[ -z $1 ]]; then
```
pkubatrh: Would be better if the contents of the argument were used to change to the directory specified instead. Will be useful if there are any additional tests added in the future.

test/run
```diff
@@ -45,34 +45,50 @@ run_s2i_build() {
...
+  fi
   git init
   git config user.email "build@localhost" && git config user.name "builder"
   git add -A && git commit --no-gpg-sign -m "Sample commit"
   popd >/dev/null
 }
 
 run_test_application() {
-  docker run --user=100001 --rm --cidfile=${cid_file} $1 ${IMAGE_NAME}-testapp
+    if [[ -z $2 ]]; then
```
pkubatrh: Same thing over here.

test/run
```diff
@@ -68,10 +71,13 @@ prepare() {
 }
 
 run_test_application() {
-    if [[ -z $2 ]]; then
-        docker run --user=100001 --rm --cidfile=${cid_file} $1 ${IMAGE_NAME}-testapp
+    if [[ $1 == app ]]; then
+        docker run --user=100001 --rm --cidfile=${cid_file} $2 ${IMAGE_NAME}-testapp
```
pkubatrh: Ah sorry, I should have been more specific. What I meant was something along the lines of:

```docker run --user=100001 --rm --cidfile=${cid_file} $2 ${IMAGE_NAME}-test${1}```

so that we do not need any if statements at all.

kasicka: Actually I wanted to handle cases when an invalid argument was passed to the function.

pkubatrh: Right, I missed that `set -e` is not used in nodejs test script. In that case it makes sense to me.

pkubatrh:
[test]

pkubatrh:
[test]

pkubatrh:
Tests passing now, lgtm. Thanks!

PR #179, opened 2018-03-19
use local reference policy for tags

bparees:


imagestreams/nodejs-centos7.json
```diff
@@ -49,14 +55,17 @@
           "openshift.io/provider-display-name": "Red Hat, Inc.",
           "description": "Build and run Node.js 4 applications on CentOS 7. For more information about using this builder image, including OpenShift considerations, see https://github.com/sclorg/s2i-nodejs-container/blob/master/4/README.md.",
           "iconClass": "icon-nodejs",
-          "tags": "builder,nodejs",
+          "tags": "hidden,builder,nodejs",
```
omron93: Why adding "hidden" tag?

pkubatrh: It goes EOL in May

bparees: that's around the time these imagestreams would be shipping in OCP3.10.  So we're deprecating the nodejs4 centos image a little early, but the rhel one is right in line.  I can hold off if you want, but i was hoping to just get these all updated now.

omron93: When image gets EOL, this repo has to be modified anyway (version removed from `Makefile`, `<VERSION>/README.md` changed to contain deprecation warning)....

But if this change is propagated to OpenShift in similar time, I wouldn't complicate this situation and merge this PR as is...

bparees: it's going to propagate to openshift master now, but it won't ship until v3.10 ships, so i think it's ok to do this part now.  Again the only complicating factor is it removes(well, hides, it's still accessible) the centos imagestream immediately for origin users, but given that we have newer image versions available, I think that's ok.

PR #182, opened 2018-04-05
Bug 1562759 - s2i support for incremental builds

wozniakjan:
The https://bugzilla.redhat.com/show_bug.cgi?id=1562759 reported a use case with incremental builds ~where BuildConfig output ImageStreamTag was chained as an input to another BuildConfig strategy.~

Trying to reproduce the issue discovered potentially another issue in `s2i/bin/assemble` script when it tries to copy the 'source repo' from `/tmp/src/*` into `./`. Given previous build already populated files into `./`, using `rsync` instead of `mv` could fix the following problem:
```
$ oc get pods
NAME          READY     STATUS      RESTARTS   AGE
bc1-1-build   0/1       Completed   0          18s
bc2-1-build   0/1       Error       0          13s

$ oc logs bc2-1-build
---> Installing application source ...
mv: cannot move '/tmp/src/.git' to './.git': Directory not empty
error: build error: non-zero (13) exit code from docker-registry.default.svc:5000/test/is1@sha256:163868c8004d41dad4560999999481a797def73190a7a5f9191754ab029d262c
```

Adding a reference to https://github.com/openshift/jenkins/pull/565 in another s2i image, where this pattern was used in similar situation.

cc: @bparees , @gabemontero

### DIFF
--- a/8/s2i/bin/assemble
+++ b/8/s2i/bin/assemble
@@ -28,7 +28,7 @@ safeLogging () {
 
 shopt -s dotglob
 echo "---> Installing application source ..."
-mv /tmp/src/* ./
+rm -rf ./ && rsync -av /tmp/src/ .
 
 if [ ! -z $HTTP_PROXY ]; then
     echo "---> Setting npm http proxy to" $(safeLogging $HTTP_PROXY)
--- a/0.10/s2i/bin/assemble
+++ b/0.10/s2i/bin/assemble
@@ -20,7 +20,7 @@ set -e
 
 shopt -s dotglob
 echo "---> Installing application source ..."
-mv /tmp/src/* ./
+rm -rf ./ && mv /tmp/src/* ./
 
 if [ ! -z $HTTP_PROXY ]; then
 	echo "---> Setting npm http proxy to $HTTP_PROXY"
--- a/4/s2i/bin/assemble
+++ b/4/s2i/bin/assemble
@@ -28,7 +28,7 @@ safeLogging () {
 
 shopt -s dotglob
 echo "---> Installing application source ..."
-mv /tmp/src/* ./
+rm -rf ./ && mv /tmp/src/* ./
 
 if [ ! -z $HTTP_PROXY ]; then
     echo "---> Setting npm http proxy to" $(safeLogging $HTTP_PROXY)
--- a/6/s2i/bin/assemble
+++ b/6/s2i/bin/assemble
@@ -28,7 +28,7 @@ safeLogging () {
 
 shopt -s dotglob
 echo "---> Installing application source ..."
-mv /tmp/src/* ./
+rm -rf ./ && rsync -av /tmp/src/ .
 
 if [ ! -z $HTTP_PROXY ]; then
     echo "---> Setting npm http proxy to" $(safeLogging $HTTP_PROXY)


8/s2i/bin/assemble
```diff
@@ -28,7 +28,7 @@ safeLogging () {
 
 shopt -s dotglob
 echo "---> Installing application source ..."
-mv /tmp/src/* ./
+rsync -av /tmp/src/ .
```
bparees: this will leave the source in the injected location which results in two copies of it in the output image, no? we don't want that.

bparees: Also this mv pattern is in pretty much all our s2i builder images, so if we do decide to make a change, all of them should be updated.

wozniakjan: > this will leave the source in the injected location which results in two copies of it in the output image, no? we don't want that.

yes, we could also `rm -rf /tmp/src/*` after, so image contains source repo only once. And also this merges into the files previous s2i image in the "chain" left in `./`. 

Or maybe we would like to plain `rm -rf ./* && mv /tmp/src* ./` to get rid of anything the previous image in the "chain" left there?

> Also this mv pattern is in pretty much all our s2i builder images, so if we do decide to make a change, all of them should be updated.

absolutely, marking this PR as [WIP] to serve a purpose of initiating a discussion about how to solve this for all s2i images

bparees: We do want it to be possible to perform incremental builds with these images, so i do agree it needs to be solved.

As for which approach to take (clearing the destination first and doing a mv, or using rsync and then clearing the source), i don't have an informed opinion right now.

omron93: > rm -rf ./* && mv /tmp/src* ./

Is the same as to prohibit using image "in such build chain", or not?

I think if we want to support similar chains, rsync is the only option. Because if user have any non-empty directory in the source, current implementation of assemble script will fail.

wozniakjan: > Is the same as to prohibit using image "in such build chain", or not?

yes, very similar. Maybe more than prohibiting I'd say ignoring, but either way the result ends up similar

> I think if we want to support similar chains, rsync is the only option

Thanks, I will update this PR for all assemble scripts (currently, it is only for nodejs) and wait for more feedback

bparees:
@omron93 i think this is ready for another review/merge.

omron93:
> this will leave the source in the injected location which results in two copies of it in the output image, no? we don't want that.

@wozniakjan This needs to be somehow resolved. Or not?

Because similar approach should be done for all our images I would welcome more reviews. @pkubatrh Please could you take a look?

wozniakjan:
> > this will leave the source in the injected location which results in two copies of it in the output image, no? we don't want that.
> 
> @wozniakjan This needs to be somehow resolved. Or not?

I personally see these two options
1. `rm -rf /tmp/src/* && mv /tmp/src/* ./` - keep only new src files
2. `rsync -av /tmp/src/ .` - merge old and new src files

So if it is not desired to merge src files, we can do the 1. But I'd prefer if the decision was made based on some feedback from s2i images users or designers, they probably know the best what suits the current needs.

bparees:
> rm -rf /tmp/src/* && mv /tmp/src/* ./ - keep only new src files

??

did you mean to do a cp(not mv) first and then an rm -rf /tmp/src after?

that would be my preferred solution since it's most consistent w/ what the image currently does.

rsync and then rm -rf /tmp/src would also be ok.

Either way I don't think we should leave the content in /tmp when we're done.

wozniakjan:
> did you mean to do a cp(not mv) first and then an rm -rf /tmp/src after?

yes, meant to copy that text from previous comments and accidentally wrote nonsense - `rm -rf ./* && mv /tmp/src/* ./` is what I meant there

> that would be my preferred solution since it's most consistent w/ what the image currently does.

thanks, I will update the PR

omron93:
Wow! Now I'm totally confused :-)

>We do want it to be possible to perform incremental builds with these images

From that I thought we want to support build chains. I'm affraid that with `rm -rf ./*` it would be quite confusing. It means we state that users can do build chains (~ use image built by s2i as base image for another s2i build) but silently it won't have any effect!

So IMHO one of this should happen:
- it should be properly documented that build chains aren't possible, so that just images released by us can be used for s2i builds
- or support build chains. That means use `rsync -av /tmp/src/ . && rm -rf /tmp/src/*` in assemble scripts.

But we should admit that we are still talking mainly about OpenShift usage. So you know more how users use s2i builds and what we want to support. @wozniakjan @bparees Opinions?

bparees:
@omron93 chained builds have nothing to do w/ s2i.  This is a chained build:
https://docs.openshift.org/latest/dev_guide/builds/advanced_build_operations.html#dev-guide-chaining-builds

>  I'm affraid that with rm -rf ./* it would be quite confusing. It means we state that users can do build chains (~ use image built by s2i as base image for another s2i build) but silently it won't have any effect!

It would still have an effect.  the saved-artifacts would still be available for reuse, they would be in /tmp/ and then the assemble script could consume them.

incremental(not chained!) builds have nothing to do w/ preserving the applications working directory.

Here is what an incremental build is supposed to do:

1) extract some content from a previous version of the image (what content is up to the save-artifacts script)
2) s2i will make that content available to the assemble script at build time

So @wozniakjan is correct that clearing out "./" can make sense for an incremental build, to ensure you have a clean slate and only reuse the artifacts you explicitly choose to reuse from what save-artifacts pulled down.  If you want to reuse the source code from the previous image, your save-artifacts script can make that possible, but it should be a deliberate choice.

So I think I am in favor of `m -rf ./ && rsync /tmp/src/ .`

Perhaps looking at an image that does support incremental builds will help clarify things:

content to reuse:
https://github.com/openshift-s2i/s2i-wildfly/blob/master/12.0/s2i/bin/save-artifacts#L10

How the content gets reused at assemble time:
https://github.com/openshift-s2i/s2i-wildfly/blob/master/12.0/s2i/bin/assemble#L10-L15

wozniakjan:
Thanks @bparees  for clarifying the use-case and terminology. I apologize for confusing everyone with my initial message talking about chaining BuildConfigs by one's output ImageStreamTag for other's input ImageStreamTag, I will be more careful with the terminology.

omron93:
> chained builds have nothing to do w/ s2i. This is a chained build:
https://docs.openshift.org/latest/dev_guide/builds/advanced_build_operations.html#dev-guide-chaining-builds

Good to know. Thanks.

> Here is what an incremental build is supposed to do:
>
>    1. extract some content from a previous version of the image (what content is up to the save-artifacts script)
>    2. s2i will make that content available to the assemble script at build time

How does this relate to this image? I thought that to support incremental builds `save-artifacts` script is needed - this image doesn't have such script.

> incremental(not chained!) builds have nothing to do w/ preserving the applications working directory.

But the linked bug is not about incremental build. Am I wrong?
User is using "resulting image as input for another s2i build", so is `rm -rf ./` what user wants? Everything what he done in the first build would be removed before the second one...

bparees:
> How does this relate to this image? I thought that to support incremental builds save-artifacts script is needed - this image doesn't have such script.

users can provide their own save-artifacts script to be used w/ any existing image, and then do an incremental build w/ it, just like they can provide their own assemble+run scripts to turn any image into an s2i builder image.

That said, I do think I was thinking about this wrong.  Here's how incremental is supposed to work:

first build:
take input builder image A, produce app image B

subsequent builds:
take app image B, extract artifacts, provide artifacts along w/ new source to build using image A

I think what @wozniakjan did when trying to test incremental is:

first build:
take input builder image A, produce app image A' (replaced image A with a new version of A that included his source)

subsequent build:
take input builder image A', extract artifacts from app image A', attempt to build w/ artifacts+source+builder image A', this will fail since A' contains source in /tmp/src

So the issue here is that @wozniakjan "corrupted" his builder image by adding /tmp/src to it.  That's not what you would normally do for an incremental build (you should not be modifying the builder image, just the app image).

@wozniakjan can you confirm/does that make sense?

If so, i think we can close this PR.

> But the linked bug is not about incremental build. Am I wrong?

the linked bug is not directly related to incremental build but @wozniakjan ran into further problems using incremental build while exploring that bug/attempting to reproduce.

> Everything what he done in the first build would be removed before the second one...

not everything, just the src.  presumably the dependencies that were downloaded would still be in the image (assuming they are not downloaded to /tmp/src, which normally they would not be since we put all the deps under /opt/app-root/src normally).  But again i think this is moot, I don't think we need this PR unless what I said above about @wozniakjan's workflow is incorrect.

wozniakjan:
> first build:
> take input builder image A, produce app image A' (replaced image A with a new version of A that included his source)
> 
> subsequent build:
> take input builder image A', extract artifacts from app image A', attempt to build w/ artifacts+source+builder image A', this will fail since A' contains source in /tmp/src
>
> So the issue here is that @wozniakjan "corrupted" his builder image by adding /tmp/src to it. That's not what you would normally do for an incremental build (you should not be modifying the builder image, just the app image).

I think this is very accurate description of what I did and also what I thought the [BZ](https://bugzilla.redhat.com/show_bug.cgi?id=1562759) was reporting, my apologies for confusing everyone on multiple fronts here when trying to reproduce/fix this.

To be sure, I will just rephrase what I wanted to achieve in my words - have two `BuildConfigs`, both with strategy `source`. First `BuildConfig` has a `from` field s2i `ImageStreamTag` and output user defined `ImageStreamTag` A'. Second `BuildConfig` has a `from` field `ImageStreamTag` A' that contains `.git` and other source files from previous build. That was my understanding of the reported BZ.

And final question, is the scenario as described not supported or is it only missing some step (perhaps the `safe-artifacts` script)?

omron93:
> To be sure, I will just rephrase what I wanted to achieve in my words - have two BuildConfigs, both with strategy source. First BuildConfig has a from field s2i ImageStreamTag and output user defined ImageStreamTag A'. Second BuildConfig has a from field ImageStreamTag A' that contains .git and other source files from previous build. That was my understanding of the reported BZ.
>
> And final question, is the scenario as described not supported or is it only missing some step (perhaps the safe-artifacts script)?

Exactly. I understood it similarly (I hope:-)).

In other words, app image built by s2i is used as "source" image for second s2i build. It makes sense IMHO because the second build could be only "extension" of the application resulting from first build (it not only about reusing artifacts from a previous build). In that case `rsync -av /tmp/src/ . && rm -rf /tmp/src/*` should be used.
Who can decide the question if we want to support this?

bparees:
@wozniakjan you're totally correct that it was what the BZ reporter was doing.  I should have followed more closely and sooner realized the path we were on.

So with that in mind, I think @omron93 nailed the potential use case (extending an application image) and next question:  Do we want to support this.

My feeling is no, it's not critical to support this.  If you really want to extend an application image, it's unlikely you just want to run the same assemble script again, you probably want to do some other/new assemble logic.  And if you're introducing/replacing the assemble logic, you can rm -rf /tmp/src yourself as part of that.

So i'm not saying we shouldn't do this, but it doesn't seem particularly critical.

that said, the decision should be made by the SCL team as they own the image and the experience around it.

omron93:
We can discuss this during our SCL container meeting. @bparees Is this a priority issue or can it wail till next regular meeting in 13 days?

bparees:
It can definitely wait.

Ben Parees | OpenShift

wozniakjan:
@omron93 perfect, thank you! I think it is for the best if all related PRs are handled by you as an expert on the matter so I am closing this in favor of your future PR.

bparees:
>  @bparees Is anything else from application source internally analyzed by openshift?

nothing comes to mind.

omron93:
> (and we don't analyze the .git folder within the image anyway)

Maybe I don't understand again:-D
So how its possible that @wozniakjan solved the reported problem by using `rsync` which "copies" `.git/` folder into image? What I'm missing?

omron93:
> @omron93 the problem isn't that something is missing, problem is that something is already there from a previous step and a part of assemble script with mv now fails with non-zero exit code, rsync would have merged the two contents instead, but that isn't the core of the issue

Understand. But the "symptom" of reported issue is that labels are not updated during s2i build to reflect used source. I thought that it's connected to inability to move `.git/` which could be fixed by rsync. (I know that other folders are affected too) @wozniakjan So my assumption about this was wrong?

wozniakjan:
> But the "symptom" of reported issue is that labels are not updated during s2i build to reflect used source.

that part actually never reproduced, using the "stock" s2i images, the scenario didn't work at all, using fixed s2i images with `rsync`, the labels were populated correctly

omron93:
> that part actually never reproduced, using the "stock" s2i images, the scenario didn't work at all, using fixed s2i images with rsync, the labels were populated correctly

@wozniakjan So problem was that the second build wasn't successful and no image was produced? Not that build succeeded but labels were wrong?

wozniakjan:
> @wozniakjan So problem was that the second build wasn't successful and no image was produced? Not that build succeeded but labels were wrong?

yes, this PR was trying to address the `assemble` script issue only which was discovered during the effort to reproduce the original BZ (and testing with the fix as presented earlier, the labels were correct)

PR #193, opened 2018-09-21
OKD rebranding and sclorg move

liangxia:
`docs.openshift.org` to `docs.okd.io`
`github.com/openshift/nodejs-ex` to `github.com/sclorg/nodejs-ex`


liangxia:
@pkubatrh PTAL

omron93:
@liangxia Looks good. Thanks.

PR #200, opened 2019-02-08
Add Integration tests to be run in a remote cluster

hhorak:
The idea is to have an easy way to run tests in an existing cluster, that might be remote via `make test-openshift-remote-cluster` or directly via `test/run-openshift-remote-cluster`.


hhorak:
[test-openshift]

hhorak:
There is some problem with nodejs-ex: https://github.com/sclorg/nodejs-ex/issues/212

phracek:
This is covered by https://github.com/sclorg/s2i-nodejs-container/pull/249

Feel free to reopen it, in case it is valid.

PR #218, opened 2019-10-30
add ubi license label to rhel7 nodejs-10

pkubatrh:


10/Dockerfile.rhel7
```diff
@@ -44,6 +44,7 @@ LABEL summary="$SUMMARY" \
       com.redhat.component="rh-${NAME}${NODEJS_VERSION}-container" \
       name="rhscl/$NAME-$NODEJS_VERSION-rhel7" \
       version="1" \
+      com.redhat.license_terms="https://www.redhat.com/en/about/red-hat-end-user-license-agreements#UBI" \
```
phracek: Sorry, some of our images use #rhel7 and some #UBI like here https://github.com/sclorg/mysql-container/blob/master/8.0/Dockerfile.rhel7#L32

I would use the same license terms.

pkubatrh: That is expected, as not all of the images are released under the UBI license, but rhscl/nodejs-10 is:
https://access.redhat.com/containers/#/registry.access.redhat.com/ubi7/nodejs-10

phracek:
LGTM.

pkubatrh:
Thanks for review! Merging

PR #219, opened 2019-11-21
add ubi8-based nodejs-12 image

pkubatrh:


pkubatrh:
[test-openshift]

12/Dockerfile
```diff
@@ -0,0 +1,72 @@
...
+# This image will be initialized with "npm run $NPM_RUN"
+# See https://docs.npmjs.com/misc/scripts, and your repo's package.json
+# file for possible values of NPM_RUN
+# Description
+# Environment:
+# * $NPM_RUN - Select an alternate / custom runtime mode, defined in your package.json files' scripts section (default: npm run "start").
+# Expose ports:
+# * 8080 - Unprivileged port used by nodejs application
+
+ENV NODEJS_VERSION=10 \
```
phracek: This is intended? There should be `12`, I guess

12/Dockerfile.fedora
```diff
@@ -0,0 +1,78 @@
...
+# This image will be initialized with "npm run $NPM_RUN"
+# See https://docs.npmjs.com/misc/scripts, and your repo's package.json
+# file for possible values of NPM_RUN
+# Description
+# Environment:
+# * $NPM_RUN - Select an alternate / custom runtime mode, defined in your package.json files' scripts section (default: npm run "start").
+# Expose ports:
+# * 8080 - Unprivileged port used by nodejs application
+
+ENV NODEJS_VERSION=10 \
```
phracek: The same here.

12/Dockerfile.fedora
```diff
@@ -0,0 +1,78 @@
+FROM registry.fedoraproject.org/f29/s2i-base:latest
```
phracek: Really `f29`. What about `f30`?

12/Dockerfile.rhel7
```diff
@@ -0,0 +1,74 @@
...
+# This image will be initialized with "npm run $NPM_RUN"
+# See https://docs.npmjs.com/misc/scripts, and your repo's package.json
+# file for possible values of NPM_RUN
+# Description
+# Environment:
+# * $NPM_RUN - Select an alternate / custom runtime mode, defined in your package.json files' scripts section (default: npm run "start").
+# Expose ports:
+# * 8080 - Unprivileged port used by nodejs application
+
+ENV NODEJS_VERSION=10 \
```
phracek: The same here.

pkubatrh:
>Sorry, why we have a file called 12/.exclude-rhel7?

So that we do not build a rhel7 version of the image (as only rhel8 is being added), same for centos7 and fedora.

phracek:
> So that we do not build a rhel7 version of the image (as the only rhel8 is being added), same for centos7 and fedora.

Then I would prefer to remove Dockerfiles for centos7, fedora, and rhel7. Existing Dockerfiles for this OS are a bit confusing if they are not build at all.

pkubatrh:
>Then I would prefer to remove Dockerfiles for centos7, fedora, and rhel7. 

That would remove the git history we are working to preserve by doing it like this.

We could only add rhel8 Dockerfiles in this PR and leave the current history in 10/ to be added later, but that way we just make more work for us as we need to do the "git mv, git cp" dance multiple times.
It might also result in more confusion during adding rhel7 versions, as in https://github.com/sclorg/mariadb-container/pull/114

What I could do is take a look at the Fedora Dockerfile at least and move it to a Fedora version that actually uses nodejs v.12.

pkubatrh:
Maybe I could modify the sources (change version from available 10 to unavailable 12) so that it at least does not build?

phracek:
I got the note about git history. thanks

But if we use nodejs-10 version in directory 12, we should remember to change it later on as soon as nodejs-12 is ready for the nodejs-12 container.

phracek:
@pkubatrh As soon as you change versions to 12 as we discussed on the meeting and .exclude files stays, we can merge it this pull request.

pkubatrh:
Moved all Dockerfiles in `12/` to `NODEJS_VERSION=12`. @phracek PTAL

The Fedora version *should* build, but leaving it excluded in this PR and will follow-up in a different PR.

10/s2i/bin/assemble
```diff
@@ -100,9 +100,7 @@ else
 	if ! mountpoint $NPM_CACHE; then
 		echo "---> Cleaning the npm cache $NPM_CACHE"
 		#As of npm@5 even the 'npm cache clean --force' does not fully remove the cache directory
-		# instead of $NPM_CACHE* use $NPM_CACHE/*.
-		# We do not want to delete .npmrc file.
-		rm -rf "${NPM_CACHE:?}/"
+		rm $NPM_CACHE* -rf
```
phracek: Please change it back again.
`rm -rf $NPM_CACHE* -rf` removes also ~/.npmrc file which should be present for NPM_REGISTRY settings.

pkubatrh: Thanks for spotting this. Must have missed this during the rebase. Will fix and force push once more.

pkubatrh:
Force push to add missing fix for npm cache removal

pkubatrh:
[test-openshift]

pkubatrh:
Thanks for review! Tests are all green, so merging

PR #221, opened 2019-11-28
Do not remove .npmrc file

phracek:
function $NPM_CACHE* removes also
.npmrc file which is not needed. We need
them if `npm config set registry` is called.

Required: container-common-scripts#129


pkubatrh:
@phracek Change looks good, thanks! Can you update the common submodule as well, so we can get rhel8 CI to pass?

phracek:
@pkubatrh The submodule is up2date, but waiting on merge https://github.com/sclorg/container-common-scripts/pull/129

PR #226, opened 2019-12-23
Backport few changes previously done in RHOAR container

hhorak:
Most notable changes:
 * npx is used for nodemon execution
 * GIT_COMMITTER_NAME and GIT_COMMITTER_EMAIL variables are set
 * npm cache is cleared by rm -rf .npm (this is commented out, since we've actually had a test for specifically this, that cache is not removed)
 * npm install is run once more when starting the node
 * run npm run build when a build script is present
 * run yarn if  is set, although we do not ship yarn in our images

As a site-effect, these changes were done:
 * Match also singular in the test_incremental_build
 * Print the overall test result after the cleanup messages, so it is more visible.
 * Use supported fedora 31 for v10 Fedora Dockerfile
 * Move install-time steps to nodejs_container_install file (fixes also a wrong /usr/bin/nodemon symlink in case of SCLs)

More details in the commit messages in particular commits.

### DIFF
--- a/10/s2i/bin/assemble
+++ b/10/s2i/bin/assemble
@@ -135,8 +135,11 @@ else
 
 fi
 
-echo "---> Cleaning up npm cache"
-rm -rf .npm
+# TODO: Do we need cache or not actually?
+# based on the following commit, it doesn't seem so:
+# https://github.com/sclorg/s2i-nodejs-container/commit/0e9b5cbb2fe719c447ed3a96fd534f6b2e2ca888
+# echo "---> Cleaning up npm cache"
+# rm -rf .npm
 
 # Fix source directory permissions
 fix-permissions ./
--- a/10/s2i/bin/assemble
+++ b/10/s2i/bin/assemble
@@ -38,6 +38,10 @@ mv /tmp/src/* ./
 # Fix source directory permissions
 fix-permissions ./
 
+if [ "$DEV_MODE" == true ] ; then
+  set -x
+fi
+
 if [ ! -z $HTTP_PROXY ]; then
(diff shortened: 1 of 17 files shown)


hhorak:
[test-openshift]

10/s2i/bin/assemble
```diff
@@ -78,19 +83,37 @@ if [ -z "$NODE_ENV" ]; then
   fi
 fi
 
-if [ "$NODE_ENV" != "production" ]; then
+echo -e "Current git config"
+git config --list
+
+if [ -n "$YARN_ENABLED" ]; then
```
phracek: Implementation of `YARN_ENABLED` is fine. But we will have the same problem with NEXUS_REGISTRY.
I would prefer to add to the condition `NPM_MIRROR` also `yarn config set registry` or define a new variable called YARN_MIRROR.

12/s2i/bin/assemble
```diff
@@ -78,19 +83,37 @@ if [ -z "$NODE_ENV" ]; then
   fi
 fi
 
-if [ "$NODE_ENV" != "production" ]; then
+echo -e "Current git config"
+git config --list
+
+if [ -n "$YARN_ENABLED" ]; then
```
phracek: The same here with YARN registry

phracek:
Thanks for the PR! Nicely done with couple of comments.

phracek:
@hhorak Is this still valid? If so, I would prefer to update it. If not feel free to close it.

hhorak:
This is ancient, let's forget about this forever :)

PR #230, opened 2020-02-27
set SCL env variables on build

pkubatrh:


pkubatrh:
WIP as tests are still needed. I am thinking three cases:
1. check if `podman exec ... env` and `podman run ... env` have the same variables exported (those that matter)
2. check if `podman run ... env` matches what `scl enable $SCL env` would provide (tricky, need a way to disable SCL)
3. check if one can call npm in Dockerfile when building on top of the nodejs image

pkubatrh:
This should be now ready for review
[test][test-openshift]

pkubatrh:
[test-openshift]

PR #232, opened 2020-04-16
Run express.js test suite as part of the tests

khardix:
This patch installs and runs express.js test suite on top of the base node image.

The goal is to check if the containerized Node expresses the same behavior as the upstream one. This should help ensure that even when i.e. patching Node to use older OpenSSL than officially supported, the result is still behaving as expected by an independent project.

Changes made in best-effort way; please kindly notify me if I reinvented wheel anywhere within the patch :slightly_smiling_face:.

### DIFF
--- a/test/run
+++ b/test/run
@@ -89,9 +89,11 @@ prepare_express_repo() {
     --config advice.detachedHead="false" \
     --branch "${EXPRESS_REVISION}" --depth 1 \
     'https://github.com/expressjs/express.git' "$1"
+  pushd "$1" >/dev/null || return
   for key in "${!gitconfig[@]}"; do
-    git -C "$1" config --local "$key" "${gitconfig[$key]}"
+    git config --local "$key" "${gitconfig[$key]}"
   done
+  popd "$1" >/dev/null || return
 }
 
 prepare() {
--- a/test/run
+++ b/test/run
@@ -6,6 +6,9 @@
 # IMAGE_NAME specifies a name of the candidate image used for testing.
 # The image has to be available before this script is executed.
 #
+# EXPRESS_REVISION specifies which express.js branch or tag should be tested;
+# by default it uses the latest released version as reported by
+# `npm show express version`.
 
 THISDIR=$(dirname ${BASH_SOURCE[0]})
 
@@ -13,6 +16,7 @@ source "${THISDIR}/test-lib.sh"
 
 test -n $IMAGE_NAME \
   -a -n $VERSION
+readonly EXPRESS_REVISION="${EXPRESS_REVISION:-$(docker run --rm "${IMAGE_NAME}" -- npm show express version)}"
 
 test_dir="$(readlink -f $(dirname ${BASH_SOURCE[0]}))"
 image_dir="$(readlink -f ${test_dir}/..)"
@@ -24,6 +28,13 @@ s2i_args="--pull-policy=never "
 
 # TODO: This should be part of the image metadata
 test_port=8080
+
+# Common git configuration
+readonly -A gitconfig=(
+  [user.name]="builder"
+  [user.email]="build@localhost"
+  [commit.gpgsign]="false"
+)
 
 info() {
   echo -e "\n\e[1m[INFO] $@...\e[0m\n"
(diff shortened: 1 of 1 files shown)


hhorak:
[test]

test/run
```diff
@@ -6,13 +6,16 @@
...
+# EXPRESS_REVISION specifies which express.js branch or tag should be tested.
 
 THISDIR=$(dirname ${BASH_SOURCE[0]})
 
 source "${THISDIR}/test-lib.sh"
 
 test -n $IMAGE_NAME \
   -a -n $VERSION
+# Change when new LTS version comes out
+readonly EXPRESS_REVISION="${EXPRESS_REVISION:-4.17.1}"
```
hhorak, line 18: I'd suggest to use the latest released version, that can be retrieved by `npm show express version`.

hhorak:
Only a comment about hard-coding the `express` version that is IMO not necessary. Otherwise looks good to me.

hhorak:
CentOS 7 test fails because of `package nss_wrapper is not installed` error -- it used to be part of the mirror repos (http://vault.centos.org/7.6.1810/sclo/x86_64/rh/common/), but it is not any more. I think this needs to be fixed, because otherwise I don't know where else to get that package.

hhorak:
[tes]t

hhorak:
[test]

khardix:
[test]

test/run
khardix: Replaced with `$(npm show express version)` as default, but keeping the `EXPRESS_REVISION` variable to allow overriding it.

khardix:
[test]

khardix:
It seems that npm is only installed in the container. In order to utilize `npm show express version`, I would then have to move the express preparation (git clone and config) into the running container; no obvious way how to do this jumps out at me. Any hints? :)

hhorak:
[test]

hhorak:
>  In order to utilize `npm show express version`, I would then have to move the express preparation

You can run only the `npm show...` in the container:
```
...
--branch "${EXPRESS_REVISION:-$(docker run --rm centos/nodejs-12-centos7 'npm show express version')}" --depth 1 \ 
...
```

khardix:
> You can run only the `npm show...` in the container

This one seems to work, thanks :slightly_smiling_face:  I have adjusted the express version query and squashed the fixup commits, so this should be ready for review again.

hhorak:
[test]

test/run
```diff
@@ -57,25 +68,54 @@ run_s2i_build_proxy() {
...
+  git commit -m "Sample commit"
+}
+
+prepare_express_repo() {
+  git clone \
+    --config advice.detachedHead="false" \
+    --branch "${EXPRESS_REVISION}" --depth 1 \
+    'https://github.com/expressjs/express.git' "$1"
+  for key in "${!gitconfig[@]}"; do
+    git -C "$1" config --local "$key" "${gitconfig[$key]}"
```
phracek, line 93: Option `-C` does not exist in our testing `git` version. I would prefer to switch into `--work-dir` option.
```suggestion
    git --work-dir "$1" config --local "$key" "${gitconfig[$key]}"
```
https://git-scm.com/docs/git#Documentation/git.txt---work-treeltpathgt

phracek: See error from tests:

https://gist.github.com/rhscl-bot/c1714c8f217a97bca53924c6c165a626#file-s2i-nodejs-container-179-sh-L3443

```bash
Running express.js test suite
Cloning into '/root/sources/test/express.js'...
Unknown option: -C
usage: git [--version] [--help] [-c name=value]
           [--exec-path[=<path>]] [--html-path] [--man-path] [--info-path]
           [-p|--paginate|--no-pager] [--no-replace-objects] [--bare]
           [--git-dir=<path>] [--work-tree=<path>] [--namespace=<name>]
           <command> [<args>]
```

hhorak:
[test] [keep-machine]

test/run
khardix: Yeah, I noticed and used pushd/popd to change to the git root to avoid messing with extra CLI arguments… and then GitHub had a hiccup and the commit remained unpushed on my machine :slightly_smiling_face: Pushing it now.

khardix:
[test] [keep-machine]

hhorak:
[test]

Issue #212, opened 2019-06-21
Add the ability to use npm ci instead of npm install


pauldburton:
It would be awesome if this image had the ability to execute `npm ci` instead of `npm install` on the node 10+ images.

karangarg45:
Second this..Yesterday our whole app broke because it downloaded latest dependencies for some packages since we had caret symbol in front of our dependency version.

phracek:
@pauldburton @karangarg45 Hi folks, thanks for the issue. I have read the `npm ci` https://docs.npmjs.com/cli/ci.html#description. As I don't know the `NPM` in details, can you please help me understand what is a difference? What about to create a PR and update `s2i-nodejs-container`? Does it make sense to you?

karangarg45:
@phracek so any project which uses npm has a package.json file which specifies the version of every dependency that needs to be installed while running npm install.
In package.json there is an option to tell npm to install the latest version of the dependency every time by specifying some symbols like tilde(~) caret(^) depending if you want the latest of the patch version or latest of the minor version. 
So suppose i have some thing like this in my package.json file
`"package-a":"~1.2.3"`
When i run npm install then npm checks if there is a latest version with pattern `1.2.x`. So if a version `1.2.4` exists in npm it will just directly install that version. This is the reason we have a package-lock.json file which basically is used to tell the npm which versions to install. But this package-lock.json is only read when npm ci is used instead of npm install. It also does a lot of other things. This SO answer pretty much gives a detailed explanation
https://stackoverflow.com/a/53325242/7454011

phracek:
@karangarg45 It makes sense to me know.

phracek:
@karangarg45 @pauldburton Can you please make a pull request on it for all supported versions? 10 and 12?

phracek:
We do not have resources to fix it. The issue is 'stale'. Feel free to re-open it and file PR.

PR #236, opened 2020-04-22
Adding ability to run npm ci by default

aksgupta14:
Making `npm ci` to run by default and if it fails for any reason we will run `npm install` as a backup, this will prevent build from failure for developers who doesn't use or update  `package-lock.json`.

This is the one way and another way is to use variable `RUN_NPM_CI` like we use `NODE_ENV`, but that would require existing developers to update their `build_template` and for new developers who are not aware of this env variable won't be able to use this approach.

Please refer below docs for more information:
[npm ci](https://docs.npmjs.com/cli-commands/ci.html)
[npm install](https://docs.npmjs.com/cli-commands/install.html)

#212

### DIFF
--- a/10/s2i/bin/assemble
+++ b/10/s2i/bin/assemble
@@ -81,12 +81,30 @@ fi
 if [ "$NODE_ENV" != "production" ]; then
 
 	echo "---> Building your Node application from source"
+
+	# We will be using npm CI by default for more information please refer to
+	# https://github.com/sclorg/s2i-nodejs-container/issues/212
+	echo "---> Installing dependencies with npm ci"
+	npm ci || \
+	# npm ci will fail if certain conditions aren't met as mentioned in
+	# https://docs.npmjs.com/cli-commands/ci.html#description
+	echo "---> npm ci failed, installing dependencies with npm install"; \
 	npm install
 
 else
 
 	echo "---> Installing all dependencies"
-	NODE_ENV=development npm install
+
+	# We will be using npm CI by default for more information please refer to
+	# https://github.com/sclorg/s2i-nodejs-container/issues/212
+	echo "---> Installing dependencies with npm ci"
+	npm ci || \
+	# npm ci will fail if certain conditions aren't met as mentioned in
+	# https://docs.npmjs.com/cli-commands/ci.html#description
+	echo "---> npm ci failed, installing dependencies with npm install"; \
+	npm install
+
+	NODE_ENV=development
 
 	#do not fail when there is no build script
 	echo "---> Building in production mode"
--- a/12/s2i/bin/assemble
+++ b/12/s2i/bin/assemble
@@ -81,12 +81,29 @@ fi
 if [ "$NODE_ENV" != "production" ]; then
 
 	echo "---> Building your Node application from source"
-	npm install
 
+	# We will be using npm CI by default for more information please refer to
+	# https://github.com/sclorg/s2i-nodejs-container/issues/212
+	echo "---> Installing dependencies with npm ci"
+	npm ci || \
+	# npm ci will fail if certain conditions aren't met as mentioned in
+	# https://docs.npmjs.com/cli-commands/ci.html#description
+	echo "---> npm ci failed, installing dependencies with npm install"; \
+	npm install
 else
 
 	echo "---> Installing all dependencies"
-	NODE_ENV=development npm install
+
+	# We will be using npm CI by default for more information please refer to
+	# https://github.com/sclorg/s2i-nodejs-container/issues/212
+	echo "---> Installing dependencies with npm ci"
+	npm ci || \
+	# npm ci will fail if certain conditions aren't met as mentioned in
+	# https://docs.npmjs.com/cli-commands/ci.html#description
+	echo "---> npm ci failed, installing dependencies with npm install"; \
+	npm install
+
+	NODE_ENV=development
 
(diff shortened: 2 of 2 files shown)


phracek:
[test]

10/s2i/bin/assemble
```diff
@@ -81,12 +81,30 @@ fi
...
+	# We will be using npm CI by default for more information please refer to
+	# https://github.com/sclorg/s2i-nodejs-container/issues/212
+	echo "---> Installing dependencies with npm ci"
+	npm ci || \
+	# npm ci will fail if certain conditions aren't met as mentioned in
+	# https://docs.npmjs.com/cli-commands/ci.html#description
+	echo "---> npm ci failed, installing dependencies with npm install"; \
+	npm install
+
+	NODE_ENV=development
```
phracek, line 107:
```suggestion

```

10/s2i/bin/assemble
```diff
@@ -81,12 +81,30 @@ fi
...
-	NODE_ENV=development npm install
+
+	# We will be using npm CI by default for more information please refer to
+	# https://github.com/sclorg/s2i-nodejs-container/issues/212
+	echo "---> Installing dependencies with npm ci"
+	npm ci || \
+	# npm ci will fail if certain conditions aren't met as mentioned in
+	# https://docs.npmjs.com/cli-commands/ci.html#description
+	echo "---> npm ci failed, installing dependencies with npm install"; \
+	npm install
```
phracek, line 105:
```suggestion
	NODE_ENV=development npm install
```

12/s2i/bin/assemble
```diff
@@ -81,12 +81,29 @@ fi
...
+	# We will be using npm CI by default for more information please refer to
+	# https://github.com/sclorg/s2i-nodejs-container/issues/212
+	echo "---> Installing dependencies with npm ci"
+	npm ci || \
+	# npm ci will fail if certain conditions aren't met as mentioned in
+	# https://docs.npmjs.com/cli-commands/ci.html#description
+	echo "---> npm ci failed, installing dependencies with npm install"; \
+	npm install
+
+	NODE_ENV=development
```
phracek, line 106:
```suggestion

```

12/s2i/bin/assemble
```diff
@@ -81,12 +81,29 @@ fi
...
-	NODE_ENV=development npm install
+
+	# We will be using npm CI by default for more information please refer to
+	# https://github.com/sclorg/s2i-nodejs-container/issues/212
+	echo "---> Installing dependencies with npm ci"
+	npm ci || \
+	# npm ci will fail if certain conditions aren't met as mentioned in
+	# https://docs.npmjs.com/cli-commands/ci.html#description
+	echo "---> npm ci failed, installing dependencies with npm install"; \
+	npm install
```
phracek, line 104:
```suggestion
	NODE_ENV=development npm install
```

aksgupta14:
@phracek that is correct will push updated PR. Also is there any changes that I have to make for the test to be successful?

10/s2i/bin/assemble
aksgupta14: I have updated my PR.

phracek:
[test]

aksgupta14:
@phracek can the test failure be because of `npm ci` throwing error?

phracek:
@aksgupta14 
The failure is:

```bash
---> Installing dependencies with npm ci
[91mnpm ERR! cipm can only install packages with an existing package-lock.json or npm-shrinkwrap.json with lockfileVersion >= 1. Run an install with npm@5 or later to generate it, then try again.
[0m[91m
npm ERR! A complete log of this run can be found in:
npm ERR!     /opt/app-root/src/.npm/_logs/2020-04-27T08_37_33_712Z-debug.log
[0m---> npm ci failed, installing dependencies with npm install
```

aksgupta14:
@phracek after going through the test cases, I was able to figure out the issue. The issue appears only with incremental build and the reason being `npm ci` removes all packages inside node_modules even though it fails, now when `npm install` is run it installs all the packages that is 121 packages which make it equal to the previous build, hence incremental build errors out by giving error `ERROR Incremental build failed: both builds installed 121 packages`.

Now when we will use `npm ci` it will always install the all packages irrespective of the packages already installed which will make this test case irrelevant. can you please advice as to how I should proceed with updating test cases?

phracek:
I will investigate it a bit more after a couple days. Please be patient.

gmahomarf:
@phracek Are there any updates on this feature?

mhdawson:
This PR seems to make the change for Node.js versions 10 and 12 which are no longer supported.  While I can see the change with a working fallback makes sense to me I'd suggest it should only be made for the currently supported LTS versions. 

This is probably because it was submitted many years ago. At this point I think we should likely only do for 16 so that we minimize risk to people already using the containers.

helio-frota:
hi folks, friendly ping here about this PR.

We found a situation that maybe this feature can help us:
https://github.com/nodeshift-starters/devfile-sample/issues/10#issue-1318826158

This error is not happening when using the version 14.

mhdawson:
@aksgupta14 are you still willing/interested in getting this feature added?  I know this is quite old but looking to see how we move it forward.

phracek:
@aksgupta14 The new versions also are present. Please rebase and update it.

PR #237, opened 2020-05-04
Fix Dockerfile.fedora

hhorak:
To make the Fedora dockerfiles working again (in Fedora koji as well), we need two changes:
* use correct ENV (typo `s/VERSION/NODEJS_VERSION/`)
* not install nodemon in the Dockerfile (can be moved to the assemble script for Fedora, before the nodemon is available as RPM)

This PR also updates the Dockerfile to use the latest fedora releases in FROM.


pkubatrh:
Just wondering - what is the reason for moving the nodemon installation process from Dockerfile to the assemble script?

hhorak:
> Just wondering - what is the reason for moving the nodemon installation process from Dockerfile to the assemble script?

If it is in the Dockerfile, we ship it in the image, but since we don't have it packaged as RPM yet, we effectively ship random bits from the Internet in the image, which is probably not ok from Fedora perspective. Moving it to the assemble script is similar to what we do in case of python's PIP (it gets updated in the assemble script if the user needs it). So, the effect is pretty much the same, only the image is not shipping something not coming from koji.

phracek:
What about to wait, till nodejs-nodemon is packaged in Fedora.
I would split it into two PRs. One for NODEJS_VERSION and Fixing `FROM` tag
and another one to add nodejs-demon RPM package.

hhorak:
> not ok from Fedora perspective.

Ah, forgot to mention -- koji does not have Internet working during build, so the current version is not building in koji. That was the original motivation to do the change actually.

hhorak:
> What about to wait, till nodejs-nodemon is packaged in Fedora.

The package review was created 3 years back. Not sure how long we will wait for having nodemon packaged properly in RPMs. We have the problem in the Dockerfile now, it's not building in koji as it is now, so waiting does not seem to me like a viable option.

hhorak:
> I would split it into two PRs. One for NODEJS_VERSION and Fixing `FROM` tag
> and another one to add nodejs-demon RPM package.

I can do that if you think it's worth it.

pkubatrh:
>Ah, forgot to mention -- koji does not have Internet working during build, so the current version is not building in koji

Ah, makes sense now that you mention it. Can you add the reason to the commit message as well? Otherwise I am ok with merging as is

hhorak:
> Can you add the reason to the commit message as well?

Sure, my fault it was not there from the beginning.

hhorak:
Commit message improved now.

pkubatrh:
lgtm now, thanks!

PR #245, opened 2020-07-13
Update imagestreams

yselkowitz:


phracek:
It seems ok. 

[test][test-openshift]

yselkowitz:
cc @hhorak @lholmquist @gabemontero please review this as a sample of the discussed restructuring.

imagestreams/nodejs-centos.json
```diff
@@ -0,0 +1,147 @@
...
+          "openshift.io/provider-display-name": "Red Hat, Inc.",
+          "description": "Build and run Node.js 12 applications on UBI 8. For more information about using this builder image, including OpenShift considerations, see https://github.com/sclorg/s2i-nodejs-container/blob/master/12/README.md.",
+          "iconClass": "icon-nodejs",
+          "tags": "builder,nodejs",
+          "version": "12",
+          "sampleRepo": "https://github.com/sclorg/nodejs-ex.git"
+        },
+        "from": {
+          "kind": "DockerImage",
+          "name": "registry.redhat.io/ubi8/nodejs-12:latest"
```
gabemontero, line 44: So an OKD vs. OCP thought @yselkowitz  .... when you take the next step and start updating openshift/library with this stuff, was your thought that the community.yaml would be updated to point to this file, the official.yaml, or both?

I ask because @vrutkovs recently updated samples operator to allow for installing a set of samples, and I'm pretty sure one of the stipulations was that none of the content would come from `registry.redhat.io` since that requires credentials, and in particular, credentials which are not needed as part of the install pull secret for installing OKD (correct me if I'm wrong @vrutkovs ).  

Whereas with OCP the registry.redhat.io/TBR creds are suppose to be in the install pull secret.

So if all that lines up, then we cannot include `12-ubi8`  in the community/OKD version.  Only in the official/OCP version.

vrutkovs: +1, community imagestreams should not have images from `registry.redhat.io` as those require a pull secret.

It seems there is `registry.access.redhat.com/ubi8/nodejs-12:latest` - could we use that?

hhorak: That's good point, for the community variant of the imagestreams, we should refer to `registry.access.redhat.com/ubi8/...`. All the UBI images should be available in this non-authenticated registry as well as in the authenticated registry `registry.redhat.io`.

imagestreams/nodejs-rhel.json
```diff
@@ -0,0 +1,147 @@
...
+        "from": {
+          "kind": "DockerImage",
+          "name": "registry.redhat.io/rhscl/nodejs-12-rhel7:latest"
+        },
+        "referencePolicy": {
+          "type": "Local"
+        }
+      },
+      {
+        "name": "12",
```
hhorak, line 70: Shouldn't this tag rather link to `12-ubi7 (`"kind": "ImageStreamTag",`) to not duplicate data if possible?

hhorak:
Except the registry change that @vrutkovs suggested and a note about using `ImageStreamTag` for the `10` and `12`, LGTM.

imagestreams/nodejs-rhel.json
yselkowitz: I considered it, it would be cleaner, but the OpenShift documentation seems to imply that using ImageStreamTag makes it "floating" (iow possibly subject to future change).  @gabemontero any thoughts here?

imagestreams/nodejs-centos.json
yselkowitz: Ok, fixed that locally, will repush after a resolution to the ImageStreamTag comment.

imagestreams/nodejs-rhel.json
gabemontero: you could have both a a 12-ubi7 and 12 tag pointing to the same image

12 is needed for backward compatibility since that 10 already exists ... 12-ubi7 could be the "new form"

And then when a new image comes around after this ... say v16 ... it can just be 16-ubiwhatever

yselkowitz: Based on that, I left this as it was, let me know if you really think otherwise.

gabemontero:
LGTM

imagestreams/nodejs-rhel.json
hhorak: Ok, I'm fine with how it is now.

hhorak:
LGTM

PR #252, opened 2020-08-26
Update common sub-module

hhorak:


pkubatrh:
Note that tests will still likely fail due to: https://github.com/sclorg/container-common-scripts/pull/168

hhorak:
Fedora failing because f33/s2i-base still includes rawhide repos, and there is an issue with the key. A new f33/s2i-base build is not yet available after F33 branched, but the test should be fixed right after that. Merging now to make the imagestream tests working at least at this point.

PR #264, opened 2021-02-01
Fix link issue

phracek:
This commit fixes issue with the missing examples directory in dist-git.

https://bugzilla.redhat.com/show_bug.cgi?id=1900746#c9

### DIFF
--- a/test/examples
+++ b/test/examples
@@ -1,0 +1,1 @@
+../examples


phracek:
[test]

pkubatrh:
Since we are already linking to the root `test` directory from all the versioned dirs it might make more sense to move the symlinks there. Unless we have some reason to have versioned examples.

phracek:
The symlink is done from the `test` directory.

[test]

PR #265, opened 2021-02-01
Do not use parent dir

phracek:
Examples are symlinked into test directory


phracek:
[test]

phracek:
@pkubatrh WDYT?

pkubatrh:
LGTM, though would have been better if this was part of #264 but that one is already merged.

phracek:
Next time it will be better;)

PR #267, opened 2021-02-03
Add DEV_MODE tests and fix testcase results

phracek:
Add dev_mode tests that were missing from migration to run_all_tests.

The commit also checks the return value of each testcase.
In case the testcase does not exist, then it failed.

### DIFF
--- a/test/run
+++ b/test/run
@@ -25,18 +25,28 @@ test_nodemon_removed
 test_npm_cache_cleared
 test_npm_tmp_cleared
 kill_test_application
-test_dev_mode_app_true_development
-test_dev_mode_app_false_production
+test_dev_mode_true_development
+test_dev_mode_false_production
 "
 
-TEST_LIST_DEV="\
+TEST_LIST_NODE_ENV="\
 test_run_app_application
 test_connection
 test_nodemon_present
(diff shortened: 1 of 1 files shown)


phracek:
@fila43 Please take a look at it. Thanks.

[test]

test/run
```diff
@@ -25,18 +25,28 @@ test_nodemon_removed
...
 test_nodemon_present
 test_npm_cache_exists
 kill_test_application
-test_dev_mode_app_true_development
-test_dev_mode_app_false_development
+test_dev_mode_true_development
+test_dev_mode_false_development
+"
+
+TEST_LIST_DEV_MODE="\
```
fila43, line 42: TEST_LIST_DEV_MODE and TEST_LIST_DEV_MODE are same is it right? Otherwise LGTM

phracek: Not really. `TEST_LIST_NODE_ENV` contains a test `test_dev_mode_false_development` at the end but `TEST_LIST_DEV_MODE` contains a test `test_dev_mode_false_production` at the end.

fila43: ok thanks so we can merge it

PR #270, opened 2021-04-01
Make sure module is properly reset before attempt to install

hhorak:
This reqiures https://github.com/sclorg/s2i-base-container/pull/206 be merged first.

### DIFF
--- a/10/Dockerfile.rhel8
+++ b/10/Dockerfile.rhel8
@@ -49,7 +49,7 @@ LABEL summary="$SUMMARY" \
       help="For more information visit https://github.com/sclorg/s2i-nodejs-container" \
       usage="s2i build <SOURCE-REPOSITORY> ubi8/$NAME-$NODEJS_VERSION:latest <APP-NAME>"
 
-RUN yum -y module enable nodejs:$NODEJS_VERSION && \
+RUN yum -y module reset nodejs && yum -y module enable nodejs:$NODEJS_VERSION && \
     INSTALL_PKGS="nodejs npm nodejs-nodemon nss_wrapper" && \
     ln -s /usr/lib/node_modules/nodemon/bin/nodemon.js /usr/bin/nodemon && \
     yum remove -y $INSTALL_PKGS && \


hhorak:
[test]

hhorak:
> This reqiures [sclorg/s2i-base-container#206](https://github.com/sclorg/s2i-base-container/pull/206) be merged first.

Maybe this PR does not actually need to wait for the above (which is blocked by CI issues now), because once a change is merged in s2i-base repo, all depended community containers should get rebuild.. so if we can depend on that, merging this even now should be possible.. Still, letting @pkubatrh or @phracek to speak because I might misinterpret how the automatic community builds work.

hhorak:
[test]

10/Dockerfile.rhel8
```diff
@@ -49,7 +49,7 @@ LABEL summary="$SUMMARY" \
       help="For more information visit https://github.com/sclorg/s2i-nodejs-container" \
       usage="s2i build <SOURCE-REPOSITORY> ubi8/$NAME-$NODEJS_VERSION:latest <APP-NAME>"
 
-RUN yum -y module enable nodejs:$NODEJS_VERSION && \
+RUN yum -y module reset nodejs && yum -y module enable nodejs:$NODEJS_VERSION && \
```
phracek, line 52: The code itself looks good. Tests are passing. But I don't follow the code.

Can you please describe it?
1) install and by this pull request reset nodejs module. What the reset does mean? Any docu?
2) remove nodejs package https://github.com/sclorg/s2i-nodejs-container/blob/master/10/Dockerfile.rhel8#L55
3) install nodejs package https://github.com/sclorg/s2i-nodejs-container/blob/master/10/Dockerfile.rhel8#L56

Would it be possible to add `reset` option also to https://github.com/sclorg/s2i-nodejs-container/blob/master/14/Dockerfile.rhel8#L52 version `14`?

phracek (changes requested):

10/Dockerfile.rhel8
hhorak: Reset is necessary for switching module streams. So, in case there is already some nodejs module installed (thus enabled), we must first call `dnf module reset` and only then enable it. In case of version 14, the Nodejs Dockefile is built on top of s2i-core, not s2i-base, so we're sure there is no nodejs installed in the parent image and don't need to run `dnf module reset` in that case.

hhorak: I've put the comment to the commit message as well.

phracek:
CI is being stucked. Let's try one more.

[test]

phracek:
[test]

phracek:
[test]

PR #269, opened 2021-03-19
Nodejs 14 minimal image

pkubatrh:

PR #273, opened 2021-04-27
add buildpacks metadata to build images

pkubatrh:
Based on: https://github.com/buildpacks/spec/blob/main/platform.md#build-image


pkubatrh:
Should not affect tests in any way, but lets run them to make sure

[test][test-openshift]

mhdawson:
@pkubatrh can we also include in the minimal images as well ? Or is that covered somewhere else?

pkubatrh:
>can we also include in the minimal images as well ? Or is that covered somewhere else?

We can, but it was my understanding that this change was for build images. Would the minimal image be more of a run image (or whatever it is called in buildstacks)?

pkubatrh:
[test]

mhdawson:
@pkubatrh the buildpacks use both a build and a run image, the metadata needs to be on both.

pkubatrh:
@mhdawson Makes sense. Looking at the [metadata spec for run images](https://github.com/buildpacks/spec/blob/main/platform.md#run-image) it seems like the "must" metadata parts are slightly different - the `CNB_` environment variable definitions are missing.
I will add the definitions for the runtime minimal image to the other PR: https://github.com/sclorg/s2i-nodejs-container/pull/269

mhdawson:
@pkubatrh I've not tested without the CNB_ env variables in the run image (since I only had one I used it for both).  

Will we easily be able to respin the images if it turns out we need them?

PR #274, opened 2021-05-12
Missing git and openssl-devel makes problem in some npm modules

hhorak:
Originaly reported in https://bugzilla.redhat.com/show_bug.cgi?id=1950018

It was agreed on the Node.js stakeholders meeting that with nodejs minimal
image we don't need to keep the build nodejs 14 container that small any more.


hhorak:
[test]

pkubatrh:
lgtm

PR #278, opened 2021-06-18
chore:  hide ubi7 versions of Node.js.

lholmquist:
* This hides the ubi7 versions on Node so they will not appear as options for a user to pick.

 * Motivation:  The drop down list in the Dev Console is showing ubi7 versions of Node before it shows the ubi8.  This seems to be an issue not just for Node.js but for the other runtime options.  Since we(the node team) would like to only promote the use if ubi8(the latest as of this writing), hiding ubi7 from the list the "quick fix" option for the time being"

//cc @yselkowitz

### DIFF
--- a/imagestreams/nodejs-centos.json
+++ b/imagestreams/nodejs-centos.json
@@ -54,7 +54,7 @@
           "openshift.io/provider-display-name": "Red Hat, Inc.",
           "description": "Build and run Node.js 14 applications on UBI 7. For more information about using this builder image, including OpenShift considerations, see https://github.com/sclorg/s2i-nodejs-container/blob/master/14/README.md.",
           "iconClass": "icon-nodejs",
-          "tags": "builder,nodejs",
+          "tags": "builder,nodejs,hidden",
           "version": "14",
           "sampleRepo": "https://github.com/sclorg/nodejs-ex.git"
         },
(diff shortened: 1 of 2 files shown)


hhorak:
Since this change will be available in the default imagestreams only in the upcoming OpenShift versions that focus on RHEL-8 anyway, it makes sense to me to promote ubi8 images only by hiding ubi7, so this change makes sense to me. We'll need to check with rest of the team whether they're ok with this change.

It would be also good to hear from @yselkowitz whether it seems ok from OpenShift perspective.

yselkowitz:
Not my call to make.  cc @gabemontero @adambkaplan

gabemontero:
yeah this is fine 

just to be clear, the ubi7 imagestreams will still be installed in ocp/okd per their definitions in openshift/library

for example, of the ocp/rhel version, https://github.com/openshift/library/blob/master/official.yaml#L149

but yes, hiding them from the console's drop down will help discourage their use

adambkaplan (approved):
LGTM

lholmquist:
> just to be clear, the ubi7 imagestreams will still be installed in ocp/okd per their definitions in openshift/library

Yea,  that is expected.  we just wanted to hide them to promote ubi8 instead

lholmquist:
Now that this has been merged,  is there a next step that i need to do to make sure it makes it into Openshift?  or is this somehow automatically imported?

@yselkowitz @gabemontero @pkubatrh

yselkowitz:
No, you're set.  It was already automatically synced to openshift/library:

https://github.com/openshift/library/commit/933ea6de05b9aab4160d3fff5dca4f65f611999d

From there, it will be (manually) synced into cluster-samples-operator some time during 4.9 development.

lholmquist:
@yselkowitz thanks for the info!

PR #280, opened 2021-07-15
Add support for Testing Farm CI

phracek:
This pull request adds support for testing s2i-nodejs-container in Testing Farm

Supported scenarios are:

Fedora
CentOS7
CentOS7 with OpenShift

How does it work?
As soon as a pull request is created, when the owner or member writes the comment [test], then Testing Farm is scheduled.

### DIFF
--- a/.github/workflows/centos7-openshift-tests.yaml
+++ b/.github/workflows/centos7-openshift-tests.yaml
@@ -1,0 +1,188 @@
+name: CentOS7-openshift-tests@TF
+
+on:
+  issue_comment:
+    types:
+      - created
+jobs:
+  build:
+    # This job only runs for '[test]' pull request comments by owner, member
+    name: Schedule test on Testing Farm service for CentOS7 - Openshift
+    runs-on: ubuntu-20.04
+    if: |
+      github.event.issue.pull_request
+      && github.event.comment.body == '[test]'
+      && contains(fromJson('["OWNER", "MEMBER"]'), github.event.comment.author_association)
+    outputs:
+      REQ_ID: ${{steps.sched_test.outputs.REQ_ID}}
+      SHA: ${{steps.sha.outputs.SHA}}
+    steps:
+      - name: Get pull request number
+        id: pr_nr
+        run: |
+          PR_URL="${{ github.event.comment.issue_url }}"
+          echo "::set-output name=PR_NR::${PR_URL##*/}"
+
+      - name: Checkout repo
+        uses: actions/checkout@v2
+        with:
+          ref: "refs/pull/${{ steps.pr_nr.outputs.PR_NR }}/head"
+
+      - name: Get sha
+        id: sha
+        run: |
+          echo "::set-output name=SHA::$(git rev-parse HEAD)"
+
+      - name: Create status check to pending
+        id: pending
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          cat << EOF > pending.json
+          {
+            "sha": "${{steps.sha.outputs.SHA}}",
+            "state": "pending",
+            "context": "Testing Farm - CentOS7 - OpenShift 3",
+            "target_url": "http://artifacts.dev.testing-farm.io/${{ steps.sched_test.outputs.req_id }}/pipeline.log"
+          }
+          EOF
+          echo "https://api.github.com/repos/$GITHUB_REPOSITORY/statuses/${{steps.sha.outputs.SHA}}"
+          curl -X POST -H "Authorization: Bearer $GITHUB_TOKEN" -H "Accept: application/vnd.github.v3+json" \
+            https://api.github.com/repos/$GITHUB_REPOSITORY/statuses/${{steps.sha.outputs.SHA}} \
+            --data @pending.json
+          echo "::set-output name=GITHUB_REPOSITORY::$GITHUB_REPOSITORY"
+
+      - name: Schedule a test on Testing Farm forn CentOS7 - OpenShift 3
+        id: sched_test
+        run: |
+          apt update && apt -y install curl jq
+          cat << EOF > request.json
+          {
+            "api_key": "${{ secrets.TF_PUBLIC_API_KEY }}",
+            "test": {"fmf": {
+              "url": "https://github.com/sclorg/sclorg-testing-farm",
+              "ref": "main"
+            }},
+            "environments": [{
+              "arch": "x86_64",
+              "os": {"compose": "CentOS-7"},
+              "variables": {
+                "REPO_URL": "$GITHUB_SERVER_URL/$GITHUB_REPOSITORY",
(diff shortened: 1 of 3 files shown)


.github/workflows/centos7-openshift-tests.yaml
```diff
@@ -0,0 +1,188 @@
...
+        id: sha
+        run: |
+          echo "::set-output name=SHA::$(git rev-parse HEAD)"
+
+      - name: Create status check to pending
+        id: pending
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          cat << EOF > pending.json
```
hhorak, line 41: Some pieces would deserve some comments, like what is pending.json file actually?

.github/workflows/centos7-openshift-tests.yaml
```diff
@@ -0,0 +1,188 @@
...
+          {
+            "sha": "${{steps.sha.outputs.SHA}}",
+            "state": "pending",
+            "context": "Testing Farm - CentOS7 - OpenShift 3",
+            "target_url": "http://artifacts.dev.testing-farm.io/${{ steps.sched_test.outputs.req_id }}/pipeline.log"
+          }
+          EOF
+          echo "https://api.github.com/repos/$GITHUB_REPOSITORY/statuses/${{steps.sha.outputs.SHA}}"
+          curl -X POST -H "Authorization: Bearer $GITHUB_TOKEN" -H "Accept: application/vnd.github.v3+json" \
```
hhorak, line 50: Another comment might explain what do we authorize for

.github/workflows/centos7-openshift-tests.yaml
```diff
@@ -0,0 +1,188 @@
...
+                "REPO_NAME": "$GITHUB_REPOSITORY",
+                "PR_NUMBER": "${{ steps.pr_nr.outputs.PR_NR }}",
+                "OS": "centos7",
+                "TEST_NAME": "test-openshift"
+              }
+            }]
+          }
+          EOF
+          curl ${{ secrets.TF_ENDPOINT }}/requests --data @request.json --header "Content-Type: application/json" --output response.json
+          echo "::set-output name=REQ_ID::$(jq -r .id response.json)"
```
hhorak, line 80: It's not clear what happens when the response includes something not expected. Maybe some error handling might improve further potential debugging.

.github/workflows/centos7-openshift-tests.yaml
```diff
@@ -0,0 +1,188 @@
...
+  running:
+    needs: build
+    name: Check running tests on Testing Farm service
+    runs-on: ubuntu-20.04
+    outputs:
+      REQ_ID: ${{steps.req_sha.outputs.REQ_ID}}
+      SHA: ${{steps.req_sha.outputs.SHA}}
+      GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+    steps:
+      - name: Check if REQ_ID and SHA exists
```
hhorak, line 104: I might miss something, but what happens if one of them does not exist? will the error be readable enough?

hhorak (commented):
Few comments regarding the maintainability of the code (adding comments, checking the output, etc), but from the design perspective, these yaml files do not seem to include anything specific for the nodejs image, so I can imaging those to be shared across the images. For example the whole workflows directory or *.yaml files separately could be symlinks to the common sub-module?

.github/workflows/centos7-openshift-tests.yaml
phracek: This is JSON file used to file a job into Testing Farm. I do not want to use the same file in the GitHub action.
But comment would be more better for the other teammates. I will definitely add the comments.

.github/workflows/centos7-openshift-tests.yaml
phracek: I will try to investigate it a bit more. Maybe the same error handling like at the end would be nice.

.github/workflows/centos7-openshift-tests.yaml
phracek: Both of this values are taken from GitHub Actions environment variables. I will update link to them and add the comment. Thanks

phracek:
> For example the whole workflows directory or *.yaml files separately could be symlinks to the common sub-module?

I will try to 'google' it a bit more if there is no possibility to have these files in the `sclorg` organization or maybe a symlink to `container-common-scripts` repository. Good idea, though

phracek:
https://github.community/t/plans-to-support-global-workflows-in-github-repository/17899

phracek:
@hhorak Unfortunately, we can not use symlinks. See error here: https://github.com/phracek/s2i-nodejs-container/actions/runs/1045165679 . But with normal file it is working properly, but with some kind of failure caused by my sync https://github.com/phracek/s2i-nodejs-container/actions/runs/1045177090

phracek:
I will investigate it a bit more with https://docs.github.com/en/actions/learn-github-actions/sharing-workflows-with-your-organization

phracek:
I have created a repository called https://github.com/sclorg/.github which should be used for all our container images. Files in containers will not be duplicated.

.github/workflows/centos7-openshift-tests.yaml
phracek: Testing Farm always returns ID which is not null.
Format is like:
```json
{"id": "8c9fd9ac-ce64-47bc-ac94-41fa977121b2", "test": {"fmf": {"url": "https://gitlab.cee.redhat.com/platform-eng-core-services/sclorg-tmt-plans", "ref": "master", "path": ".", "name": "rhel7"}, "script": null, "sti": null}, "state": "new", "environments": [{"arch": "x86_64", "os": {"compose": "RHEL-7.9-Released"}, "pool": null, "variables": {"REPO_URL": "https://github.com/sclorg/s2i-ruby-container", "REPO_NAME": "s2i-ruby-container", "PR_NUMBER": "316", "OS": "rhel7", "TEST_NAME": "test-openshift-4"}, "artifacts": null, "settings": null, "tmt": null}], "notification": null, "created": "None", "updated": "None"}
```

The next action called "running" checks for `state` value from response.json file https://github.com/sclorg/s2i-nodejs-container/pull/280/files#diff-5eeb4c94288868dbab08fb59bcbf4f13a23990675a6cd33145bae53150e2fa42R133

phracek:
@hhorak Based on the page https://docs.github.com/en/actions/learn-github-actions/sharing-workflows-with-your-organization I have found a way how to create a shared GitHub actions for all sclorg organization. See PR: https://github.com/sclorg/.github/pull/1.

As soon as https://github.com/sclorg/.github/pull/1 is merged, I will try to do it on one our container which does not have `.github/workflows` directory.

hhorak:
This looks interesting, thanks for looking at the de-duplication :)

phracek:
Closing this PR, so we will use workflow template files from https://github.com/sclorg/.github repository

PR #293, opened 2021-10-21
Support cvp pipeline

phracek:
This commit adds support for CVP pipeline.

It adds `test/test-openshift.yaml` file into upstream. As soon as https://github.com/sclorg/container-common-scripts/pull/214 is merge. The file `test/test-openshift.yaml` will be symlink to `common/test-openshift.yaml`

The corresponding downstream dist-git will contain also `test/test-openshift.yaml` file with generated version done by https://github.com/sclorg/container-workflow-tool/pull/50


phracek:
[test-all]

PR #298, opened 2021-11-18
Update imagestreams for RHEL 8.5

yselkowitz:
- Update imagestreams for RHEL 8.5
- Use groupified apiVersion

rh-nodejs12 goes out of support in December.


pkubatrh:
LGTM, thanks!

PR #306, opened 2021-12-10
Deprecate nodejs 12

zmiklank:
Remove nodejs 12 from README.md, mark it in 12/README.md as deprecated, remove it from imagestreams and Makefile.

### DIFF
--- a/.github/workflows/build-and-push.yml
+++ b/.github/workflows/build-and-push.yml
@@ -9,7 +9,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        version: [12, 14, 16]
+        version: [14, 16]
     steps:
       - uses: actions/checkout@v2
 
--- a/README.md
+++ b/README.md
@@ -1,8 +1,6 @@
 NodeJS container images
 ====================
 
-s2i-nodejs-container 12 Quay.io status: [![Docker Repository on Quay](https://quay.io/repository/centos7/nodejs-12-centos7/status "Docker Repository on Quay")](https://quay.io/repository/centos7/nodejs-12-centos7)
-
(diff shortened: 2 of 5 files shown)


pkubatrh:
Note that only ubi7-based nodejs 12 is getting deprecated at the end of Dec 2021. ubi8-based nodejs 12 images is supported until end of April 2022

zmiklank:
@pkubatrh So the imagestream files wouldn't be affected with this deprecation at all?

pkubatrh:
@zmiklank They would be but Yaakov already removed the streams in https://github.com/sclorg/s2i-nodejs-container/pull/298

What we need to do for this image is make sure the ubi7 and centos7 versions are not built and tested anymore (by adding `.exclude.rhel7` and `.exclude.centos7` files).
Actually we can keep the removals of ubi8 imagestreams for version 12 of the image, since we should be removing them in advance anyway, but the version 12 needs to be present in the Makefile so that the ubi8 version is still buildable. The README needs to be present as well, until we eventually deprecate the ubi8 version.

pkubatrh:
If possible please separate the imagestream removals into a second commit with a commit message describing that we are removing them in advance of the planned image depracation in April 2022.

zmiklank:
@pkubatrh I have reworked the PR. Is it OK now?

pkubatrh:
Just one nitpick - please remove the quay.io message from the README as well (https://github.com/sclorg/s2i-nodejs-container/blob/master/README.md?plain=1#L4) since that is tied to the centos7 image.

zmiklank:
Thanks for pointing it out. I have updated the README.

pkubatrh:
Thanks! LGTM now
[test-all]

pkubatrh:
[test-openshift]

PR #309, opened 2021-12-21
fix setting PATH for minimal images

pkubatrh:
we cannot set both HOME and PATH using $HOME in the same ENV command


pkubatrh:
[test-all]

phracek:
[test-all]

phracek:
@pkubatrh in case of Fedora failures, can you please rebase `common`? Fix is this 
https://github.com/sclorg/container-common-scripts/pull/234

PR #310, opened 2022-01-03
run all openshift4 tests, even if one fails

zmiklank:
Showing results at the end of the testsuite needs to be supported also in the common scripts.

### DIFF
--- a/test/run-openshift-remote-cluster
+++ b/test/run-openshift-remote-cluster
@@ -13,12 +13,19 @@ source "${THISDIR}/test-lib-openshift.sh"
 source "${THISDIR}/test-lib-nodejs.sh"
 source "${THISDIR}/test-lib-remote-openshift.sh"
 
+TEST_LIST="\
+test_nodejs_s2i_container
+test_nodejs_s2i_app_ex
+test_nodejs_s2i_templates
+test_nodejs_imagestream
+"
+
 
 # change the branch to a different value if a new change in the example
 # app needs to be tested
 BRANCH_TO_TEST=master
 
-set -eo nounset
+set -u
 
 trap ct_os_cleanup EXIT SIGINT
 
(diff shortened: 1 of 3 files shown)


zmiklank:
[test-openshift]

zmiklank:
[test-openshift]

zmiklank:
[test-all]

zmiklank:
PR for container-common-scripts support is here: https://github.com/sclorg/container-common-scripts/pull/226

test/test-lib-nodejs.sh
```diff
@@ -519,6 +520,18 @@ function test_safe_logging() {
   fi
 }
 
+function save_test_result() {
```
phracek, line 523: Why you are creating a new function? The variables can be a part of `run_all_tests`. I don't follow.
What about moving the function into `container-common-scripts` so we can use it generally.

test/test-lib-nodejs.sh
```diff
@@ -532,15 +545,17 @@ function run_all_tests() {
...
-    else
-      test_msg="[FAILED]"
-      TESTSUITE_RESULT=1
-    fi
-    printf -v test_short_summary "%s %s for '%s' %s\n" "${test_short_summary}" "${test_msg}" "${APP_NAME}" "$test_case"
-    [ -n "${FAIL_QUICKLY:-}" ] && cleanup "${APP_NAME}" && return 1
+    save_test_result $TESTCASE_RESULT $APP_NAME
+  done;
+}
+function os_run_all_tests() {
```
phracek, line 551: This is not needed. we have run_all_tests function

phracek: Calling of tests can be executed by:
```bash
TEST_SET=${TESTS:-$TEST_LIST} run_all_tests "openshift-4"
```

phracek (changes requested):
Thanks for the PR. I think we do not need to create more functions. Just one is enough.

See an inspiration here and below https://github.com/sclorg/s2i-nodejs-container/blob/master/test/run#L142

phracek:
At the end of the CI logs, I don't see a list of test cases.
`ct_os_cleanup` function has to be called.

zmiklank:
> At the end of the CI logs, I don't see a list of test cases. `ct_os_cleanup` function has to be called.

The `ct_os_cleanup` function is called automatically from the `run-openshift-remote-cluster` script on exit from the shell and on SIGINT signal. It is initiated with trap builtin: `trap ct_os_cleanup EXIT SIGINT`.
The reason, why the list of TCs is not visible in the above test runs is that support in container-common-scripts is needed.
When this [1] PR is merged, the list will be visible (it is tested locally).

[1]  https://github.com/sclorg/container-common-scripts/pull/226

test/test-lib-nodejs.sh
zmiklank: I have created this new function, because the `run_all_tests` is specific for nodejs container tests. In openshift tests we do not need `$CLIENT_LIST` and run client TCs in a specific way.
But it is true, that it can be written differently. Moving a more generic version of `run_all_tests` to `container-common-scripts` seems like a good idea, I will try it.

zmiklank: PR for ccontainer-common-scripts available here: https://github.com/sclorg/container-common-scripts/pull/229

zmiklank:
@phracek, the https://github.com/sclorg/container-common-scripts/pull/229 is already merged. 
I have also created PR for updating common scripts to the newest version https://github.com/sclorg/s2i-nodejs-container/pull/317.

When it is merged this can be tested properly.

phracek:
This is not needed, you can add the latest container-common-scripts into this pull request directly.

zmiklank:
Ok, next time I will do common submodule bump in the same PR.

zmiklank:
[test-openshift]

phracek (changes requested):
Thanks for the PR!. Tests are passing which is awesome. But I would prefer to have even definitions in the main file `test/run-openshift-remote-cluster`. So the functions are visible in the main file.

test/run-openshift-remote-cluster
```diff
@@ -13,12 +13,21 @@ source "${THISDIR}/test-lib-openshift.sh"
...
+TESTSUITE_RESULT=0
 
 # change the branch to a different value if a new change in the example
 # app needs to be tested
 BRANCH_TO_TEST=master
 
-set -eo nounset
+set -u
 
 trap ct_os_cleanup EXIT SIGINT
```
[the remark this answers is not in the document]
zmiklank: IMO it is, as we do not want the test suite to exit when the first test case fails.

zmiklank:
Thank you for the review.

> But I would prefer to have even definitions in the main file test/run-openshift-remote-cluster. So the functions are visible in the main file.

Aren't all testcase functions located in the `test-lib-nodejs.sh`?

phracek (approved):
LGTM. Thanks for handling.

phracek:
[test-all]

PR #320, opened 2022-03-08
migrate to the new build-and-push-action

zmiklank:


.github/workflows/build-and-push.yml
```diff
@@ -8,54 +8,17 @@ jobs:
...
-          image: ${{ steps.build-image.outputs.image }}
-          tags: ${{ steps.build-image.outputs.tags }}
-          registry: quay.io/centos7
-          username: ${{ secrets.QUAY_IMAGE_BUILDER_USERNAME }}
-          password: ${{ secrets.QUAY_IMAGE_BUILDER_TOKEN }}
-
-      - name: Print image url
-        run: echo "Image pushed to ${{ steps.push-to-quay.outputs.registry-paths }}"
+          registry: "quay.io"
+          registry_namespace: "centos7"
```
phracek, line 19: What about C9S? This repository belongs to sclorg.

.github/workflows/build-and-push.yml
```diff
@@ -8,54 +8,17 @@ jobs:
...
+          registry_username: ${{ secrets.QUAY_IMAGE_BUILDER_USERNAME }}
+          registry_token: ${{ secrets.QUAY_IMAGE_BUILDER_TOKEN }}
+          docker_context: ${{ matrix.dockerfile_path }}
+          dockerfile_path: ${{ matrix.dockerfile_path }}
+          tag: "centos7"
```
phracek, line 24: The same here. It should be mentioned in the matrix.

phracek (changes requested):
Thanks for the PR! Please address also my changes.

.github/workflows/build-and-push.yml
zmiklank: Thanks for the review.
I only rewrote the action as it was. I didn't want to add any new functionality. 

I would suggest creating other PR for adding c9s support, to test it separately, but if you really want, I can add it here, WDYT?

.github/workflows/build-and-push.yml
```diff
@@ -8,54 +8,17 @@ jobs:
   build-and-push:
     runs-on: ubuntu-latest
     strategy:
```
phracek, line 10:
```suggestion
    if: github.repository_owner == 'sclorg'
    strategy:
```

phracek (changes requested):
Thanks for the PR!. Look at my proposals.

.github/workflows/build-and-push.yml
```diff
@@ -7,55 +7,19 @@ on:
 jobs:
   build-and-push:
     runs-on: ubuntu-latest
+    if: github.repository_owner == 'sclorg'
     strategy:
+      fail-fast: false
       matrix:
-        version: [14, 16]
+        dockerfile_path: [ "14", "16" ]
```
phracek, line 14: And what bout minimal?
As far as I know, we supports `12 14 14-minimal 16 16-minimal`

phracek (changes requested):
Thanks for the PR!. Please address my changes.

.github/workflows/build-and-push.yml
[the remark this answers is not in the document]
zmiklank: This is already included.

.github/workflows/build-and-push.yml
zmiklank: They have their own action, but I can check and compare, and add them here if possible.

PR #328, opened 2022-04-11
Build push c9s and fedora to quay

phracek:
This pull request builds CentOS7, Fedora and C9S images to quay.io


phracek:
[test]

phracek:
CentOS  9 tests failed:

```
 [PASSED] for 'clients' express
 [FAILED] for 'clients' pino
 [PASSED] for 'clients' prom-client
```

zmiklank (changes requested):
Hi, thanks for the PR.

LGTM in general, however, when I looked into separate commits I noticed, that I do not understand the meaning of this one: f05e3932327a8ddc36bb7671acbfe4a36c15f958, as the commit message does not correspond with the commit changes.
Could you please look at it, or explain?

phracek:
> Hi, thanks for the PR.
> 
> LGTM in general, however, when I looked into separate commits I noticed, that I do not understand the meaning of this one: [f05e393](https://github.com/sclorg/s2i-nodejs-container/commit/f05e3932327a8ddc36bb7671acbfe4a36c15f958), as the commit message does not correspond with the commit changes. Could you please look at it, or explain?

NodeJS 16 is not supported in RHEL7 and CentOS 7. Only RHEL 8.

zmiklank:
Oh, I see, thanks.

Could you then please change the commit message to something like "do not build nodejs16 for centos7"?

zmiklank (approved):
Thanks.

pkubatrh:
Note that at this point I would be ok (in future PRs) with just removing the Dockerfile for rhel7 and centos7 versions as we are unlikely to ship new versions of images for those operating systems.

PR #345, opened 2022-07-14
Use 9.0.0 to have the packages available as 9.1.0 is not released yet

hhorak:


hhorak:
[test]

phracek (changes requested):
Add please also for `openshift-tests.yml` in the same directory.
As soon as it is done, we can merge it.

PR #350, opened 2022-08-23
[DRAFT] Build for c9s as the nodejs:18 is available in c9s now

hhorak:
The new stream is only available in the test compose, so this is a prepared PR only to be merged once the new nodejs:18 stream is available at mirrors as well.


hhorak:
Not valid any more, node.js 18 is ubi, so we don't need c8s/c9s any more.

PR #353, opened 2022-09-12
Support testing separate versions

phracek:
Testing single versions in CI.
It is valid for containers as for openshift tests.


phracek:
[test-all]

zmiklank (approved):

Issue #357, opened 2022-09-21
Node 18 Builder image "Dockerfile.c9s" does not complete build


pacostas:
Building docker image `Dockerfile.c9s` with `docker build -f Dockerfile.c9s . -t builderImage18c9s` throws below error during building process
```
The command '/bin/sh -c MODULE_DEPS="make gcc gcc-c++ git openssl-devel" &&     INSTALL_PKGS="$MODULE_DEPS nodejs npm nodejs-nodemon nss_wrapper" &&     ln -s /usr/lib/node_modules/nodemon/bin/nodemon.js /usr/bin/nodemon &&     yum install -y --setopt=tsflags=nodocs $INSTALL_PKGS &&     rpm -V $INSTALL_PKGS &&     node -v | grep -qe "^v$NODEJS_VERSION\." && echo "Found VERSION $NODEJS_VERSION" &&     yum -y clean all --enablerepo='*'' returned a non-zero code: 1
```
Trying to figure out what is the issue I believe it has to do with with not specifying nodejs version https://github.com/sclorg/s2i-nodejs-container/blob/e70cd8f58174d83b0eb8c619ee224dbf691d66f3/18/Dockerfile.c9s#L57-L63 
as on Dockerfile.c8s  https://github.com/sclorg/s2i-nodejs-container/blob/e70cd8f58174d83b0eb8c619ee224dbf691d66f3/18/Dockerfile.c8s#L56 

More specifically on line https://github.com/sclorg/s2i-nodejs-container/blob/e70cd8f58174d83b0eb8c619ee224dbf691d66f3/18/Dockerfile.c9s#L62 the command `node -v` returns 16 instead of 18

pacostas:
@phracek sure

PR #358, opened 2022-09-21
fix: specifying nodejs version to be on 18

pacostas:
This fix, specifies nodejs version to be on 18, otherwise the build image does not build as described on this issue https://github.com/sclorg/s2i-nodejs-container/issues/357

### DIFF
--- a/18-minimal/Dockerfile.rhel9
+++ b/18-minimal/Dockerfile.rhel9
@@ -51,6 +51,8 @@ LABEL summary="$SUMMARY" \
 
 # nodejs-full-i18n is included for error strings
 RUN INSTALL_PKGS="nodejs nodejs-nodemon nodejs-full-i18n npm findutils tar" && \
+    microdnf module disable nodejs && \
+    microdnf module enable nodejs:$NODEJS_VERSION && \
     microdnf -y --nodocs --setopt=install_weak_deps=0 install $INSTALL_PKGS && \
     node -v | grep -qe "^v$NODEJS_VERSION\." && echo "Found VERSION $NODEJS_VERSION" && \
     microdnf clean all && \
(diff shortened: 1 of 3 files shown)


mhdawson:
Makes sense to me, seems to be missing compared to other files.

mhdawson (approved):
LGTM

mhdawson:
My guess is that the assumption might have been that Node.js 18 was the default for RHEL 9, but it looks like that is not the case. @kasicka does it make sense to you that the default is Node.js 16 ?

pkubatrh:
It's just a leftover from copying the Dockerfile of the 16 version (which does not use modules) and not making necessary modifications for module use. LGTM. One thing though - @pacostas can you update `18/Dockerfile.rhel9` as well with the same change?
Note that the images are not yet set up to be built as part of the CI, so running it makes no sense now. Let's figure out the issues (if any) once we start officially building the images.

pacostas:
@pkubatrh  From what I see
- [ ] On rhel9/ubi9 node 18 is not available for installing it ATM. Will be available as a module on the next rhel9.1 release, so it might be necessary to specify the nodejs version as on `centos 8 stream` and `centos 9 stream` using the command `yum -y module enable nodejs:$NODEJS_VERSION`
- [ ] On rhel8/ubi8 as as on rhel9/ubi9 will be available on rhel8.7 version
- [ ] On centos-9-stream, node 18 is available but needs to specify nodejs version [This PR specifies nodejs version]
- [x] On centos-8-stream, node 18 is available and also specified. Image works.
- [x] On fedora , node 18 is available and also specified. Image works.

Updating the PR, specifying also NODEJS_VERSION and on rhel9.1 and rhel8.7

pacostas:
For the minimal images. Added also specifying the nodejs version on rhel9 images. Same pattern as on rhel8, centos 8 stream & centos 9 stream.

pkubatrh:
[test-all]

pacostas:
Above testing does not cover the changes. I guess we have to wait for an additional commit which will include the c9s, rhel9, rhel9 - minimal.

phracek (approved):
LGTM. Thank you

phracek:
The NodeJS 18 testing is covered by #362

PR #361, opened 2022-10-07
container_tests: allow sourcing ENV VARS from PR comment

zmiklank:
Tested here: https://github.com/zmiklank/s2i-nodejs-container/pull/1

Also, should this be documented somewhere?

### DIFF
--- a/.github/workflows/container-tests.yml
+++ b/.github/workflows/container-tests.yml
@@ -84,6 +84,14 @@ jobs:
         with:
           ref: "refs/pull/${{ github.event.issue.number }}/head"
 
+      - name: Obtain envvars from PR's comment
+        id: user_defined
+        shell: bash
+        run : |
+          _comment=${{ github.event.comment.body }}
+          _vars=$(echo $_comment | grep -Eo '\[ENV:.*\]' | cut -d':' -f2 | tr -d ']')
+          echo "::set-output name=VARS::${_vars}"
+
       - name: Prepare needed variables
         shell: bash
         id: vars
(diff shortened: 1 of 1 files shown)


.github/workflows/container-tests.yml
```diff
@@ -84,6 +84,14 @@ jobs:
         with:
           ref: "refs/pull/${{ github.event.issue.number }}/head"
 
+      - name: Obtain envvars from PR's comment
+        id: user_defined
+        shell: bash
+        run : |
+          _comment=${{ github.event.comment.body }}
+          _vars=$(echo $_comment | grep -Eo '\[ENV:.*\]' | cut -d':' -f2 | tr -d ']')
```
phracek, line 92: Do we allow more then one [ENV:<variable>]?

phracek, line 92: I would also check if variable [ENV:SOME_KEY:SOME_VALUE] contains both SOME_KEY and SOME_VALUE.

phracek (changes requested):
What about openshift tests, we are going to skipp them?

.github/workflows/container-tests.yml
zmiklank: Yeah, I should have documented it somewhere.
the [ENV:] can be used only once, but you can specify more env vars there. Example: [ENV:DEBUG=1;FAIL_FAST=3;TEST=test]

zmiklank: Is this really needed?
Only members or maintainers of the sclorg would be able to use it anyway. I personally would add this to documentation somewhere, so these people know how to use it.
But if you insist I will rework this.

phracek: What about if user specify `[ENV:DEBUG;FAIL_3;TEST]`?

phracek: I am thinking if the user can specify a variable that can run some weird code or even if it can damage infrastructure.
The tests can be triggered by owners or maintainers of our repositories, so may be I am too paranoid :)

zmiklank: That would not work. Should it be working? IMO not.

zmiklank: Hmm, I need to think about it.

phracek:
I would prefer to wait till testing CI is not up and running again. For sure.

zmiklank:
OK. But this change will not appear in this PR's tests results because files of the GitHub Actions are taken from master branch code.

zmiklank:
[test]

zmiklank:
Is there anything I need to change in this PR in order to merge it? @pkubatrh @phracek

.github/workflows/container-tests.yml
phracek: No Python please. This is ok.

phracek:
[test-all]

phracek (approved):
LGTM from my point of view. Let's test it.

.github/workflows/container-tests.yml
hhorak: And for the comment about documentation of semantics of the value -- we can either have a note how to define an arbitrary variable in the README.md in the root of this repo.. or as a comment here in the yaml?

zmiklank: I vote for the comment in yaml.

PR #369, opened 2022-11-07
Has the time already come to remove the EOL container images files?

zmiklank:
+ use version 14 in examples.

If the time has not come yet, then please close this PR.

CC @pkubatrh


zmiklank:
* I have created this PR now mainly because I want to solve the situation around nss_wrapper, and that would be easier for me, if the EOL version are not in the repository anymore.

zmiklank:
Fedora and RHEL9 failed on pino.

PR #374, opened 2023-01-13
use ct_init for initialization of the test environment

zmiklank:


phracek (approved):
LGTM. Thanks for the review.

PR #387, opened 2023-04-25
chore: removes pino temporarily from testing

lholmquist:
Based on our conversations regarding the flaky pino tests, this PR is for temporarily removing the module from testing.

One of the issue that is happening is being worked on in the upstream repo:  https://github.com/pinojs/pino/pull/1691. It looks like they have also found some timeout issues, which we've seen here, which look like it might be a couple weeks before those issues get looked at:  https://github.com/pinojs/pino/pull/1691#issuecomment-1518552096

### DIFF
--- a/test/run
+++ b/test/run
@@ -70,7 +70,6 @@ test_build_express_webapp
 # XXX_REVISION and XXX_REPO must be defined below
 TEST_LIST_CLIENTS="\
 test_client_express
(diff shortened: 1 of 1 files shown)


lholmquist:
[test]

phracek (approved):
LGTM. Thanks for removing. Unstable test pino

PR #388, opened 2023-05-09
use ct_init in test/run for minimal images

zmiklank:
`test/run` has already been updated with `ct_init` function. In order to run test correctly the `run-minimal` needs to be updated too.


zmiklank:
[test]

phracek (approved):
I did not hit any issues. Tests are passing. LGTM.

Issue #379, opened 2023-02-03
tests: fail fast when prepare step fails


pkubatrh:
While debugging issues seen in #377 I noticed that the `prepare` function failure exited the whole script.
While switching to `return 1` and checking for the result in the function where it was called from managed to fix the issue of getting the test cases ignored and whole test suite getting set as passing, the rest of the test case where the `prepare` function got called continued trying to test likely wrong images/configuration.
When the prepare step fails in a test case we should have some option to fail fast and not continue with the rest of the test case.

@phracek @zmiklank wdyt

zmiklank:
Since writing the `ct_run_tests_from_testset` the `FAIL_QUICKLY` has been added, and when it is set, the run_test function returns 1.
However I think this will not do much of an action, since the return only calls `ct_cleanup` which is trapped on EXIT signal. This will IMO lead to loosing the return value 1.
Also there is no place in the code where `FAIL_QUICKLY` is being set.

I agree with you, that some mechanism like this is needed. I think we should rework the `FAIL_QUICKLY` mechanism to work properly.

I can do that if you want/agree.

zmiklank:
Solved by this commit: https://github.com/sclorg/s2i-nodejs-container/commit/40d10c009c7a325311d25ec595015512609aca24.

PR #392, opened 2023-05-26
return from function or shell if preparation of the app fails

zmiklank:
- use generic check function
 - update common
 - use return instead of exit for graceful termination

### DIFF
--- a/test/run
+++ b/test/run
@@ -142,26 +142,26 @@ cid_file=$CID_FILE_DIR/$(mktemp -u -p . --suffix=.cid)
 prepare app
 echo "Testing the production image build"
 run_s2i_build
-check_result $?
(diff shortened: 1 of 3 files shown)


zmiklank:
[test-all]

zmiklank:
Most of the failures in the tests are related to the fastify client app. However all RHEL tests failed on some ansible stuff during preparation of the test environment. Need to look at it closer. It is probably not related to this PR.

zmiklank:
Fixes: https://github.com/sclorg/s2i-nodejs-container/issues/379

zmiklank:
[test]

phracek (approved):
Thanks for this pull request. I did not hit any issue.

PR #409, opened 2023-10-04
Enable build and push for v20 (minimal and full for cXs and minimal for Fedora)

hhorak:
The goal to have cXs images is to have a testing image for the upcoming version before it's released officially, so community can take a look sooner. Since v16 and v18 are already released in the supported UBI offering, there is no use case for the cXs based container images on quay.io.


phracek:
@hhorak What did you do with `common` submodule? Was there any change? I don't follow. Maybe typo?

hhorak:
@phracek Yeah, that was a mistake I overlooked, added an update of common that was not supposed to be part of this. Now it should be fixed and the change should only touch .github/ files. Thanks for spotting and sorry for unneeded issues.

phracek:
[test]

phracek (approved):

zmiklank:
Is it intended to disable c9s builds for nodejs 16 with this PR? If yes, could you please just state that in the commit message?

phracek:
> Is it intended to disable c9s builds for nodejs 16 with this PR? If yes, could you please just state that in the commit message?

@zmiklank As far as I know, as soon as new version of nodejs is released, then we do not need to build C9S, C8S images, because user can use UBI images.

I need to test syncing. As soon as you are fine with that comment, wait and I will merge it.

hhorak:
> Is it intended to disable c9s builds for nodejs 16 with this PR? If yes, could you please just state that in the commit message?

Thanks for the point, I should have explain that right away. Yes, it was deliberate as I don't see a need to update cXs images when the supported UBI based images are already released officially for v16 and v18. The commit message is fixed now.

zmiklank (approved):

zmiklank:
thanks.

Issue #308, opened 2021-12-19
Need Monorepo support


kunyan:
Monorepo is becoming more and more popular repo style in Node.js ecosystem.
It support the developers could arrange the packages as they want and link them together.

The current s2i steps looks not support monorepo. I can show some example which I met

the monorepo folders are like this

* packages/
  * foo/
    * src/
    * README.md
    * package.json
  * bar/
    * src/
    * README.md
    * package.json
* .eslintrc
* package.json

Some common devDependencies had been install to the root package.json. like eslint, typescript and others
So the `foo` and `bar` won't need to install them.

If I need build the `foo` only, For current s2i workflow, I have 2 choice.

### 1 Use git contextDir 
1. Clone the repo and set the `contextDir` to `foo`.
2. Run `npm run build`, however, some dependencies only installed in root, so we will got error

### 2. Hack the `build` scripts
1. Clone the repo from root
2. Install all dependencies, and add post install to install all dependencies in child `packages`
3. Add some env vars to make `build` script could detect which package I want to build

### Solution

I think the only way to fix this issue for now is make the `build` command to a env vars. make user could custom it(Like $NPM_RUN)

Add a $NPM_BUILD might be the most easy way to resolve it

KrisLamote:
I was kindof assuming that the NPM_RUN env var would be doing that. iow if present run an alternative build script during openshift Build and running an alternatif start script for the openshift Deploy. But this indeed doesn't seem to be the case.

zmiklank:
Hi @kunyan, do you consider this issue fixed by https://github.com/sclorg/s2i-nodejs-container/pull/424 being merged?

kunyan:
@zmiklank $NPM_BUILD is a good for some monorepo use `nx` or `lerna`

PR #424, opened 2024-03-22
Add NPM_BUILD environment variable

whenlit:
I would like to be able to specify the npm script for the build step instead of `build`, the same way you can use `NPM_RUN` to specify a different run script instead of `start`. My use case is that I want to build and deploy part of a monorepo. As it is, I have to copy the whole `assemble` file to my `.s2i` directory and change it in there.

I agree with the suggestion in issue #308 that the most appropriate name for this variable would be `NPM_BUILD`. I have not found anything online that would indicate that it is already in use for something else.


whenlit:
Hi @phracek, I hope you don't mind me mentioning you like this, just making sure that someone notices this PR in here.

phracek:
This pull request looks easy to review. What do you think @lholmquist @mhdawson?

lholmquist (approved):
LGTM. this is a small but effective addition

PR #425, opened 2024-03-27
Add support for ulimit into client part

phracek:
nodejs container already support ulimit to function `ct_test_app_dockerfile` that is used by most tests.

But client part uses function `ct_s2i_build_as_df` function and we need to enhance it.


zmiklank (approved):

phracek:
[test-all]
[test-upstream]

PR #427, opened 2024-04-11
Imagestream generator

zmiklank:
Here the problem is, that we have the minimal tags created differently than in other containers:
```
        "name": "16-ubi8-minimal"
```
vs
```
        "name": "16-minimal-ubi8"
```
@phracek please keep this in mind when reviewing. This could be an issue.


phracek (commented):
Well based on the our Container catalog, we have e.g https://catalog.redhat.com/software/containers/rhel9/nodejs-18-minimal/62e8e8ffa9a2233d923f5f1c?q=nodejs&architecture=amd64&image=6601e09c5781451e968f8dd5&container-tabs=gti
as `rhel9/nodejs-18-minimal` or `ubi9/nodejs-18-minimal`

I would prefer to have the same semantic as in Container Catalog, like 
but this change has to be sent to OpenShift 4 team so that they properly use it. Only for sure.

zmiklank:
> Well based on the our Container catalog, we have e.g https://catalog.redhat.com/software/containers/rhel9/nodejs-18-minimal/62e8e8ffa9a2233d923f5f1c?q=nodejs&architecture=amd64&image=6601e09c5781451e968f8dd5&container-tabs=gti as `rhel9/nodejs-18-minimal` or `ubi9/nodejs-18-minimal`
> 
> I would prefer to have the same semantic as in Container Catalog, like but this change has to be sent to OpenShift 4 team so that they properly use it. Only for sure.

Thanks. @fbm3307 Could you please take a look?

phracek:
@yselkowitz Sorry for bothering you again. What do you think about changing imagestreams names from `16-ubi8-minimal` to `16-minimal-ubi8`. How this could break functionality from users point of view. Thanks for any kind of response.

yselkowitz:
Changing existing imagestreamtag names would break any existing use of that tag.  If there was an actual need for a rename -- as there was when we moved from e.g. 14 to 14-ubiN -- then this could be accomplished by adding the new name and marking the old name as hidden, but the latter would have to be kept for the duration of the underlying image's lifetime.

phracek (changes requested):
Let's keep the old name so we do not break functional who uses our imagestreams.

phracek:
The ci-script PR has been merged.

[test-all]

phracek (approved):
LGTM. Thanks

PR #431, opened 2024-05-21
Update container-common-scripts to the latest.

phracek:
From the nightly build point of view we need to have this pull request merged.

See details Remove set -e and add to the end FAILED message (#369) from container-common-scripts


phracek:
[test]
[test-upstream]

PR #432, opened 2024-05-22
deprecate c8s images, as EOL for c8s is 31th of May 2024

zmiklank:
+ cleaning after centos7 deprecation


20-minimal/README.md
```diff
@@ -6,7 +6,6 @@ with the full-sized s2i-enabled Node.JS 20 image to build the application. The i
 but compared to the full-sized Node.JS 20 image it will be missing many build-time dependencies.
 Users can choose between RHEL, CentOS and Fedora based images.
 The RHEL images are available in the [Red Hat Container Catalog](https://access.redhat.com/containers/),
-the CentOS images are available on [Quay.io](https://quay.io/organization/centos7),
```
phracek, line 9: What about CentOS Stream 9 images? I guess it should be replaced by https://quay.io/organization/sclorg, right?

phracek (changes requested):
Some documentation issue.

20-minimal/README.md
zmiklank: Done.

20-minimal/README.md
```diff
@@ -321,5 +321,5 @@ See also
 --------
 Dockerfile and other sources are available on https://github.com/sclorg/s2i-nodejs-container.
 In that repository you also can find another versions of Node.js environment Dockerfiles.
-Dockerfile for CentOS is called `Dockerfile`, Dockerfile for RHEL7 is called `Dockerfile.rhel7`,
+Dockerfile for RHEL7 is called `Dockerfile.rhel7`,
```
phracek, line 324: CentOS Stream images are missing

phracek, line 324:
```suggestion
Dockerfile for CentOS Stream 9 is called `Dockerfile.c9s`, and Dockerfile for RHEL7 is called `Dockerfile.rhel7`,
```

20/README.md
```diff
@@ -249,5 +249,5 @@ See also
 --------
 Dockerfile and other sources are available on https://github.com/sclorg/s2i-nodejs-container.
 In that repository you also can find another versions of Node.js environment Dockerfiles.
-Dockerfile for CentOS is called `Dockerfile`, Dockerfile for RHEL7 is called `Dockerfile.rhel7`,
+Dockerfile for RHEL7 is called `Dockerfile.rhel7`,
```
phracek, line 252:
```suggestion
Dockerfile for CentOS Stream 9 is called `Dockerfile.c9s`, and Dockerfile for RHEL7 is called `Dockerfile.rhel7`,
```

phracek (changes requested):

20-minimal/README.md
zmiklank: Yes, I did not include them for purpose. I think people will find them even without mentioning this information explicitly.

zmiklank: When we delete the CentOS7 Dockerfiles, that are missing the suffix with distribution the naming of the Dockefiles becomes more straightforward.

zmiklank: Also when I was going through the repositories when deprecating c8s, I noticed that many repositories contain only minimal information about CentOS Stream images, so this would not be the only container image repository in this state.

phracek: It does not make sense to have Dockerfile for rhel7, Dockerfile for rhel8, and not for Centos Stream images.

It can lead to misunderstandings, why they are not present.

zmiklank: fixed.

phracek (approved):
LGTM

[test]

phracek:
[test]

PR #442, opened 2024-08-08
Add support for testing NodeJS 22 and NodeJS 22-minimal on C10S

phracek:


phracek:
Running tests is not needed. It only allows to test s2i-nodejs-container and version 22 on C10S

PR #449, opened 2024-09-16
Add helm charts tests into s2i-nodejs-container.

phracek:
Add helm charts tests into s2i-nodejs-container.

We need to check by each PR, that also helm charts are working properly.


phracek:
[test-openshift-pytest]

Issue #456, opened 2024-10-07
Test uses check_result which isn't defined


lukaszachy:
### Description

See test/run of nodejs-18-minimal. It uses `check_result $?` which isn't defined in used files.

PR #457, opened 2024-10-07
Use ct_check_testcase_result instead of check_result

lukaszachy:
The `ct_check_testcase_result` seems to be the right function to use instead of `check_result`

Fix: #456


phracek:
[test]

phracek (approved):
The failures are not caused by this pull request.  LGTM. Thanks for this fix.

PR #459, opened 2024-10-18
Fix assemble cache clean order in 16-minimal, 18-minimal

SlouchyButton:
Removal of the cache has to be the last things done in the assemble script. Calling `npm` after deleting the cache folder will cause npm to recreate it - which will fail during the image test.

Other images apparently have it correctly, for some reason this error is present only in 16-minimal and 18-minimal.

### DIFF
--- a/16-minimal/s2i/bin/assemble
+++ b/16-minimal/s2i/bin/assemble
@@ -96,6 +96,12 @@ else
 	npm prune
 
 	# Clear the npm's cache and tmp directories only if they are not a docker volumes
+	NPM_TMP=$(npm config get tmp)
+	if ! mountpoint $NPM_TMP; then
+		echo "---> Cleaning the $NPM_TMP/npm-*"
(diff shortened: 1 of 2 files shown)


SlouchyButton:
[test]

lukaszachy (commented):
LGTM (I'm afraid to click approve in the case it would unblock the merge bot)

SlouchyButton:
> I'm afraid to click approve
afaik two approvals are needed and it's not yet merging automatically, only sending email that it would.

phracek (commented):
LGTM, but is it enought to move cleaning operations before NPM_CACHE?

phracek:
@lukaszachy @SlouchyButton Automatic merger is not enabled yet;). Do worry. See send only mails about PR and what is missing. Syncing upstream -> downstream for RHEL-9.5.0 is also disabled.

phracek:
[test]

PR #467, opened 2025-01-06
Fix Fedora NodeJS container so tests passed

phracek:
NodeJS 22 Fedora container does not install nodejs package and therefore /usr/bin/node and the others are not present.

The commit is simple and fix the problem during build of NodeJS 22 and NodeJS 22-minimal container.
We do not install nodejs package that contains /usr/bin/node file and the rest.
Therefore build is failing.
This fixes the problem.


phracek:
Let's run only container tests

[test]

phracek:
Let's test it again after fix with stoping container that CID file is not exists.

[test]

phracek:
Fixing 20-minimal container and fix 'kill-test-application'

[test]

phracek:
[test]

pacostas (approved):
LGTM

jamacku:
[test]

PR #470, opened 2025-01-21
Enable building and testing NodeJS containers in RHEL10 host.

phracek:
Enable building and testing NodeJS containers in RHEL10 host.
No code is changed.


phracek:
Fedora and C9S tests are failed because of rdkafka. This should not block this pull request.

PR #477, opened 2025-03-05
Separate handling of non-default nodejs to sync Fedora 41 Dockerfiles

hhorak:
This PR requires the changes in the nodejs18 RPMs to get to the stable repo first, they're in bodhi now: https://bodhi.fedoraproject.org/updates/FEDORA-2025-e21c7819e3

The update ^^ fixes a problem that nodejs18 pulled in nodejs package as a dependency.

With that, we can use F41 as a base for all Node.js versions.

Since this sync to the newest Fedora required to rework the mechanism of "making the alternative Node.js to be the only one in the container and without version in CLI names", aka adding symlinks and testing that we actually have a correct version available, I've separated these steps into a separate script, rather than making the Dockerfile.fedora more complex. This did not seem to be required in RHEL/CXS versions, but once we put new alternative versions to RHEL 10, we might reuse this script there as well (i.e. create RHEL10/C10S dockerfile from the Fedora Dockerfile).

### DIFF
--- a/18-minimal/Dockerfile.fedora
+++ b/18-minimal/Dockerfile.fedora
@@ -1,4 +1,4 @@
-FROM quay.io/fedora/fedora-minimal:38
+FROM quay.io/fedora/fedora-minimal:41
 
 EXPOSE 8080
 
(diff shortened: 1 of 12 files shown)


phracek:
[test][test-upstream]

phracek:
Let's try one more time
[test][test-upstream]

phracek:
[test]

hhorak:
There is a re-design of nodejs packaging in Fedora, so it makes sense to get back to this once we have new packages available in Fedora.

PR #480, opened 2025-03-27
New GitHub action for testing nodejs container by test-upstream keyword.

phracek:
This commit separates container tests from container upstream tests

so we are able to test both of them and see results in Pull Request window directly.

See PR https://github.com/sclorg/tfaga-wrapper/pull/38


phracek:
This pull request does not need to run tests. It will be executed in the other testing pull request.

PR #482, opened 2025-04-22
Switch GitHub Action to use ubuntu-latest

phracek:
Switch GitHub Action to use ubuntu-latest

Switch GitHub Action to use ubuntu-latest.

The version 20.04 was deprecated and it is
not possible to run it at all


phracek:
Tests are not needed to be run. The version 20.04 is deprecated and it is not possible to run it again.

PR #484, opened 2025-05-02
Fix OpenShift tests and enable testing on shared_cluster

phracek:
This pull request enables testing NodeJS on Shared_cluster.

Also add prefix to helm charts.


phracek:
Node JS code was not changed. The fixed were only OpenShift Python tests.

PR #492, opened 2025-06-12
Fix regular expression for getting rhel name from rhel10 string

phracek:
Fix regular expression for getting rhel name from rhel10 string
The original regular returns 'rhel1' instead of 'rhel'.


phracek:
[test-all]

PR #493, opened 2025-06-18
Add nodejs fips test

tjuhaszrh:
Add two test cases to verify that if:
 * Container is in fips mode, node is also using fips mode.
* Container isnt in fips mode, node also isnt using fips mode.

Nodejs fips state is verified with: `node -e 'const crypto = require("crypto"); return crypto.getFips();` which should return the same value (1 for enabled, 0 for disabled) as cat "/proc/sys/crypto/fips_enabled".


phracek:
@tjuhaszrh The error is here:

```
Running test test_nodejs_fips_mode_off (starting at 2025-06-19 07:27:56+00:00) ... 
-----------------------------------------------
[eval]:1
(crypto=>{{const crypto = require(crypto); return crypto.getFips();}})(require('node:crypto'))
                                  ^

ReferenceError: Cannot access 'crypto' before initialization
    at [eval]:1:35
    at [eval]:1:71
    at runScriptInThisContext (node:internal/vm:209:10)
    at node:internal/process/execution:449:12
    at [eval]-wrapper:6:24
    at runScriptInContext (node:internal/process/execution:447:60)
    at evalFunction (node:internal/process/execution:87:30)
    at evalScript (node:internal/process/execution:99:3)
    at node:internal/main/eval_string:74:3

Node.js v22.15.0
```

phracek:
@tjuhaszrh Thanks for this pull request. I have several question regarding nodeJs.

What do you thinkg about this code?
```
function test_nodejs_fips_mode_off() {
  local ret_val
  local is_fips_enabled
  # Read fips mode from host in case exists
  if [[ -f /proc/sys/crypto/fips_enabled ]]; then
    is_fips_enabled=$(cat /proc/sys/crypto/fips_enabled)
  else
    # Set to 0 if not exists
    is_fips_enabled="0"
  fi
  if [[ "$is_fips_enabled" == "0" ]]; then
    echo "FIPS is disabled on host"
    echo "What is expected output in case disabled fips???"
    fips=$(docker run --rm ${IMAGE_NAME}-testapp /bin/bash -c "! node -e 'const crypto = require("crypto"); return crypto.getFips();'")
    echo "FIPS from app: '$fips'" # For me $fips is empty...
    if [[ "$fips" == "" ]]; then
      ct_check_testcase_result "0"
    else
      ct_check_testcase_result "$retval"
    fi
  else
    # What is expected behavior here in case FIPS is enabled and we test for fips is disabled. Does it make sense either?
    # Check fips mode in container as well
    if docker run --rm ${IMAGE_NAME}-testapp /bin/bash -c "cat /proc/sys/crypto/fips_enabled | grep -q 0"; then
      fips=$(docker run --rm ${IMAGE_NAME}-testapp /bin/bash -c "node -e 'const crypto = require("crypto"); return crypto.getFips();'")
      echo "FIPS from app: '$fips'"
      ct_check_testcase_result "$?"
    fi
  fi
}
```

tjuhaszrh:
[test]

phracek:
[test]

test/test-lib-nodejs.sh
```diff
@@ -478,6 +478,29 @@ function test_nodemon_present() {
...
+  # Test that nodejs behaves as expected in fips mode
+  local is_fips_enabled
+
+  # Read fips mode from host in case exists
+  if [[ -f /proc/sys/crypto/fips_enabled ]]; then
+    is_fips_enabled=$(cat /proc/sys/crypto/fips_enabled)
+  else
+    is_fips_enabled="0"
+  fi
+  if [ is_fips_enabled == "0" ]; then
```
phracek, line 491:
```suggestion
  if [[ "$is_fips_enabled" == "0" ]]; then
```

phracek: A nice typo here.

phracek (changes requested):
It is almost done. Thanks for this pull request. Great work.

phracek:
@tjuhaszrh Add to this pull request also fips https://github.com/sclorg/s2i-nodejs-container/blob/master/test/test-lib-nodejs.sh#L133
The code should be like 
```
    app|hw|fips|express-webapp|binary)
      pushd "${test_dir}/test-${1}" >/dev/null
      prepare_dummy_git_repo
      popd >/dev/null
      ;;
```

Add there also function like is here https://github.com/sclorg/s2i-nodejs-container/blob/master/test/test-lib-nodejs.sh#L44
```
run_s2i_build_fips() {
  ct_s2i_build_as_df file://${test_dir}/test-fips ${IMAGE_NAME} ${IMAGE_NAME}-testfips ${s2i_args} $(ct_build_s2i_npm_variables) $1
}
```

tjuhaszrh:
Added a simple app that in the context of a server verifies what kind of Hash algorithms and Ciphers are allowed to be used under FIPS.
* Allowed: SHA256, AES
* Not allowed: MLD, 3DES with only two keys.

phracek:
Rebased against upstream. Fixing some tests.

[test]

phracek:
Let's tests one more time 

[test]

PR #500, opened 2025-09-05
Enable nodejs24 for cXs and fix RHEL10 Dockerfiles

hhorak:

### DIFF
--- a/24-minimal/Dockerfile.c10s
+++ b/24-minimal/Dockerfile.c10s
@@ -51,8 +51,11 @@ LABEL summary="$SUMMARY" \
 # nodejs-full-i18n is included for error strings
 RUN INSTALL_PKGS="nodejs$NODEJS_VERSION nodejs-nodemon nodejs$NODEJS_VERSION-full-i18n nodejs$NODEJS_VERSION-npm findutils tar which" && \
     dnf -y --nodocs --setopt=install_weak_deps=0 install $INSTALL_PKGS && \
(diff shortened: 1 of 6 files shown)


hhorak:
[test]

hhorak:
[test]

phracek (changes requested):
Please address my proposals. Thanks.

hhorak:
> @hhorak Can you please also fix nodejs-24 and calling 'node' in Dockerfile? It should be on RHEL10 'node-24 -v' instead of 'node -v'. Thanks.

This is a real issue, but I'll fix it differently. We need to have `node` and other binaries without a version, so instead of changing the check, I think I need to use `rm -f` instead of `rm` only when removing `/usr/bin/node` and others, as those do not exist before creating the symlink.

hhorak:
[test]

hhorak:
[test]

phracek:
Let's re-run tests

[test-all]

phracek (approved):
LGTM. THanks for allowing NodeJS 24.

phracek:
[test-openshift]

Container tests looks good.

phracek:
Let's re-run ONLY container tests

[test]

phracek:
Let's get change to only containers test

[test]

phracek:
[test]

phracek:
[test-openshift][test-openshift-pytest]

hhorak:
The rate limit issue that makes CI fail might not be resolved for some time, so we agreed with @phracek off-line that it might be better to merge, so we can continue pulling it downstream.

PR #503, opened 2025-09-12
On RHEL10. node is not installed

phracek:
In Konflux pipeline it looks like /usr/bin/node, /usr/bin/npx, /usr/bin/npm does not exist.
And therefore build is failing.


phracek:
CLosing in favour of #500

PR #513, opened 2025-11-17
Removing langchain.js from testing suite due to yarn is necessary for installation

pacostas:
## Removing langchain.js from the testing suite.

This PR removes `langchain.js` from the testing suite because the yarn binary isn't available on the base images.

Until we further investigate how to properly include and install the yarn binary on the base images to test the `langchain.js` library, it's better to remove `langchain.js`. Currently, the failing test is a false alarm indication, as `Langchain` itself is not failing.

### DIFF
--- a/test/run-upstream
+++ b/test/run-upstream
@@ -26,7 +26,6 @@ test_client_faas
 test_client_cloudevents
 test_client_fastify
 test_client_pino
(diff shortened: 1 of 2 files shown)


phracek:
@pacostas Please rebase it agains master. And we can merge it.

PR #521, opened 2026-01-02
Fix Makefile like it is mentioned in container-common-scripts

phracek:
This pull request calls target `test-upstream` like the similar targets
are called in container-common-scripts/common.mk file.

Let's use the same way.


phracek:
[test-upstream]

phracek:
[test-all][test-upstream]

PR #543, opened 2026-06-04
Add distgen check to github workflows.

tjuhaszrh:
Add `sclorg/ci-actions/.github/workflows/distgen-check.yml@main` action to github workflows for both container-tests and container-upstream-tests.


phracek (approved):
LGTM

